PR #2, opened 2024-06-13
Refactor contract interaction

just-mitch:
Design for https://github.com/AztecProtocol/aztec-packages/issues/6973

### DIFF
--- a/in-progress/6973-refactor-base-contract-interaction.md
+++ b/in-progress/6973-refactor-base-contract-interaction.md
@@ -12,7 +12,7 @@ This is a refactor of the API for interacting with contracts to improve the user
 
 The refactored approach mimics Viem's API, with some enhancements and modifications to fit our needs.
 
-In a nutshell, by being more verbose in the API, we can remove a lot of complexity and make the code easier to understand and maintain; this also affords greater understanding and control over the lifecycle of their contracts and transactions.
+In a nutshell, by being more verbose in the API, we can remove a lot of complexity and make the code easier to understand and maintain; this also affords greater understanding and control over the lifecycle of contracts and transactions.
 
 Key changes:
 - the wallet is the central point of interaction to simulate/prove/send transactions instead of `BaseContractInteraction`
@@ -67,10 +67,13 @@ const paymentMethod = new SomeFeePaymentMethod(
 
 // Changes to the PXE (e.g. notes, nullifiers, auth wits, contract deployments, capsules) are not persisted.
 const { request: deployAliceAccountRequest } = await aliceWallet.simulate({
-  artifact: SchnorrAccountContract.artifact,
-  instance: aliceContractInstance,
-  functionName: deploymentArgs.constructorName,
-  args: deploymentArgs.constructorArgs,
+  // easy multicall support
+  calls: [{
+    artifact: SchnorrAccountContract.artifact,
+    instance: aliceContractInstance,
+    functionName: deploymentArgs.constructorName,
+    args: deploymentArgs.constructorArgs,
+  }],
   paymentMethod,
   // gasSettings: undefined => automatic gas estimation. the returned `request` will have the gasSettings set.
 });
@@ -112,14 +115,16 @@ const bananaCoinInstance = getContractInstanceFromDeployParams(
 );
 
 const { request: deployTokenRequest } = await aliceWallet.simulate({
-  artifact: TokenContract.artifact,
-  instance: bananaCoinInstance,
-  functionName: bananaCoinDeploymentArgs.constructorName,
-  args: bananaCoinDeploymentArgs.constructorArgs,
-  deploymentOptions: {
-    registerClass: true,
-    publicDeploy: true,
-  },
+  calls: [{
+    artifact: TokenContract.artifact,
+    instance: bananaCoinInstance,
+    functionName: bananaCoinDeploymentArgs.constructorName,
+    args: bananaCoinDeploymentArgs.constructorArgs,
+    deploymentOptions: {
+      registerClass: true,
+      publicDeploy: true,
+    },
+  }],
   paymentMethod
 })
 
@@ -133,9 +138,11 @@ const receipt = await sentTx.wait()
 
 ```ts
 const { result: privateBalance } = await aliceWallet.read({
-  contractInstance: bananaCoinInstance,
-  functionName: 'balance_of_private'
-  args: {owner: aliceWallet.getAddress()},
+  calls: [{
+    contractInstance: bananaCoinInstance,
+    functionName: 'balance_of_private'
+    args: {owner: aliceWallet.getAddress()}
+  }]
 });
 
 
@@ -276,16 +283,20 @@ export interface DeploymentOptions {
   publicDeploy?: boolean;
 }
 
-// new
-export interface UserRequest {
+export interface UserFunctionCall {
   contractInstance: ContractInstanceWithAddress;
   functionName: string;
   args: any;
   deploymentOptions?: DeploymentOptions;
-  gasSettings?: GasSettings;
-  paymentMethod?: FeePaymentMethod;
   contractArtifact?: ContractArtifact;
   functionAbi?: FunctionAbi;
+}
+
+// new
+export interface UserRequest {
+  calls: UserFunctionCall[];
+  gasSettings?: GasSettings;
+  paymentMethod?: FeePaymentMethod;
   from?: AztecAddress;
   simulatePublicFunctions?: boolean;
   executionResult?: ExecutionResult; // the raw output of a simulation that can be proven
@@ -371,14 +382,16 @@ Consider that we have, e.g.:
 
 ```ts
 {
-  contractInstance: bananaCoinInstance,
-  functionName: 'transfer',
-  args: { 
-    from: aliceAddress,
-    to: bobAddress,
-    value: privateBalance,
-    nonce: 0n
-  },
+  calls: [{
+    contractInstance: bananaCoinInstance,
+    functionName: 'transfer',
+    args: { 
+      from: aliceAddress,
+      to: bobAddress,
+      value: privateBalance,
+      nonce: 0n
+    },
+  }],
   paymentMethod
 }
 ```
@@ -417,40 +430,60 @@ function makeFunctionCall(
 
 ```
 
+#### main function calls
+
+Define a helper somewhere as:
+
+```ts
+const addMainFunctionCall: TxExecutionRequestAdapter = (
+  builder: TxExecutionRequestBuilder, call: UserFunctionCall
+) => {
+  if (!call.functionAbi) {
+    throw new Error('Function ABI must be provided');
+  }
+  builder.addAppFunctionCall(
+    makeFunctionCall(
+      call.functionAbi,
+      call.contractInstance.address,
+      call.args
+  ));
+}
+```
+
 #### class registration
 
 Define a helper somewhere as:
 
 ```ts
-export const addClassRegistration: TxExecutionRequestAdapter = (
-  builder: TxExecutionRequestBuilder, request: UserRequest
+const addClassRegistration = (
+  builder: TxExecutionRequestBuilder, call: UserFunctionCall
 ) => {
-    if (!request.contractArtifact) {
-      throw new Error('Contract artifact must be provided to register class');
-    }
+  if (!call.contractArtifact) {
+    throw new Error('Contract artifact must be provided to register class');
+  }
 
-    const contractClass = getContractClassFromArtifact(request.contractArtifact);
+  const contractClass = getContractClassFromArtifact(call.contractArtifact);
 
-    builder.addCapsule(
-      bufferAsFields(
-        contractClass.packedBytecode,
-        MAX_PACKED_PUBLIC_BYTECODE_SIZE_IN_FIELDS
-    ));
+  builder.addCapsule(
+    bufferAsFields(
+      contractClass.packedBytecode,
+      MAX_PACKED_PUBLIC_BYTECODE_SIZE_IN_FIELDS
+  ));
 
-    const { artifact, instance } = getCanonicalClassRegisterer();
+  const { artifact, instance } = getCanonicalClassRegisterer();
 
-    const registerFnAbi = findFunctionAbi(artifact, 'register');
+  const registerFnAbi = findFunctionAbi(artifact, 'register');
 
-    builder.addAppFunctionCall(
-      makeFunctionCall(
-        registerFnAbi,
-        instance.address,
-        {
-          artifact_hash: contractClass.artifactHash,
-          private_functions_root: contractClass.privateFunctionsRoot,
-          public_bytecode_commitment: contractClass.publicBytecodeCommitment
-        }
-    ));
+  builder.addAppFunctionCall(
+    makeFunctionCall(
+      registerFnAbi,
+      instance.address,
+      {
+        artifact_hash: contractClass.artifactHash,
+        private_functions_root: contractClass.privateFunctionsRoot,
+        public_bytecode_commitment: contractClass.publicBytecodeCommitment
+      }
+  ));
 }
 ```
 
@@ -460,8 +493,8 @@ Define a helper somewhere as
 
 ```ts
 
-export const addPublicDeployment: TxExecutionRequestAdapter = (
-  builder: TxExecutionRequestBuilder, request: UserRequest
+const addPublicDeployment = (
+  builder: TxExecutionRequestBuilder, call: UserFunctionCall
 ) => {
   const { artifact, instance } = getCanonicalInstanceDeployer();
   const deployFnAbi = findFunctionAbi(artifact, 'deploy');
@@ -478,6 +511,7 @@ export const addPublicDeployment: TxExecutionRequestAdapter = (
       }
   ));
 }
+
 ```
 
 #### Entrypoints implement `TxExecutionRequestComponent`
@@ -544,9 +578,6 @@ The abstract `BaseWallet` can implement:
 
 ```ts
 async getTxExecutionRequest(userRequest: UserRequest): Promise<TxExecutionRequest> {
-  if (!userRequest.functionAbi) {
-    throw new Error('Function ABI must be provided');
-  }
   if (!userRequest.gasSettings) {
     throw new Error('Gas settings must be provided');
   }
@@ -556,33 +587,26 @@ async getTxExecutionRequest(userRequest: UserRequest): Promise<TxExecutionReques
 
   const builder = new TxExecutionRequestBuilder();
 
-  // Add the "main" function call
-  builder.addAppFunctionCall(
-    makeFunctionCall(
-      userRequest.functionAbi,
-      userRequest.contractInstance.address,
-      userRequest.args
-  ));
+  for (const call of request.calls) {
+    addMainFunctionCall(builder, call);
+    if (call.deploymentOptions?.registerClass) {
+      addClassRegistration(builder, call);
+    }
+    if (call.deploymentOptions?.publicDeploy) {
+      addPublicDeployment(builder, call);
+    }
+    // if the user is giving us an artifact,
+    // allow the PXE to access it
+    if (call.contractArtifact) {
+      builder.addTransientContract({
+        artifact: call.contractArtifact,
+        instance: call.contractInstance,
+      });
+    }
+  }
 
   // Add stuff needed for setup, e.g. function calls, auth witnesses, etc.
   await userRequest.paymentMethod.adaptTxExecutionRequest(builder, userRequest);
-
-  if (userRequest.deploymentOptions?.registerClass) {
-    addClassRegistration(builder, userRequest);
-  }
-
-  if (userRequest.deploymentOptions?.publicDeploy) {
-    addPublicDeployment(builder, userRequest);
-  }
-
-  // if the user is giving us an artifact,
-  // allow the PXE to access it
-  if (userRequest.contractArtifact) {
-    builder.addTransientContract({
-      artifact: userRequest.contractArtifact,
-      instance: userRequest.contractInstance,
-    });
-  }
 
   // Adapt the request to the entrypoint in use.
   // Since BaseWallet is abstract, this will be implemented by the concrete class.
@@ -597,8 +621,8 @@ async getTxExecutionRequest(userRequest: UserRequest): Promise<TxExecutionReques
 
 ```ts
 // Used by simulate and read
-async #simulateInner(userRequest: UserRequest): ReturnType<Wallet['simulate']> {
-  const txExecutionRequest = await this.getTxExecutionRequest(initRequest);
+async #simulateInner(userRequest: UserRequest): ReturnType<BaseWallet['simulate']> {
+  const txExecutionRequest = await this.getTxExecutionRequest(userRequest);
   const simulatedTx = await this.simulateTx(txExecutionRequest, builder.simulatePublicFunctions, builder.from); 
   const decodedReturn = decodeSimulatedTx(simulatedTx, builder.functionAbi);
   return {
@@ -607,7 +631,7 @@ async #simulateInner(userRequest: UserRequest): ReturnType<Wallet['simulate']> {
     privateOutput: simulatedTx.privateReturnValues,
     executionResult: simulatedTx.executionResult,
     result: decodedReturn,
-    request: initRequest,
+    request: userRequest,
   };
 }
 ```
@@ -632,7 +656,7 @@ async simulate(userRequest: UserRequest): {
 
   const builder = new UserRequestBuilder(userRequest);
 
-  await this.#ensureFunctionAbi(builder);
+  await this.#ensureFunctionAbis(builder);
 
   if (builder.gasSettings) {
     return this.#simulateInner(builder.build());
@@ -653,16 +677,18 @@ async simulate(userRequest: UserRequest): {
   return result;
 }
 
-async #ensureFunctionAbi(builder: UserRequestBuilder): void {
-  // User can call simulate without the artifact if they have the function ABI
-  if (!builder.functionAbi) {
-    // If the user provides the contract artifact, we don't need to ask the PXE
-    if (!builder.contractArtifact) {
-      const contractArtifact = await this.getContractArtifact(builder.contractInstance.contractClassId);
-      builder.setContractArtifact(contractArtifact);
+async #ensureFunctionAbis(builder: UserRequestBuilder): void {
+  for (const call of builder.calls) {
+    // User can call simulate without the artifact if they have the function ABI
+    if (!call.functionAbi) {
+      // If the user provides the contract artifact, we don't need to ask the PXE
+      if (!call.contractArtifact) {
+        const contractArtifact = await this.getContractArtifact(call.contractInstance.contractClassId);
+        call.setContractArtifact(contractArtifact);
+      }
+      const functionAbi = findFunctionAbi(call.contractArtifact, call.functionName);
+      call.setFunctionAbi(functionAbi);
     }
-    const functionAbi = findFunctionAbi(builder.contractArtifact, builder.functionName);
-    builder.setFunctionAbi(functionAbi);
   }
 }
 
@@ -695,7 +721,7 @@ async read(userRequest: UserRequest): DecodedReturn | [] {
     builder.setGasSettings(GasSettings.default());
   }
 
-  await this.#ensureFunctionAbi(builder);
+  await this.#ensureFunctionAbis(builder);
 
   return this.#simulateInner(builder.build());
 }
@@ -710,7 +736,7 @@ async prove(request: UserRequest): Promise<UserRequest> {
     throw new Error('Execution result must be set before proving');
   }
   const builder = new UserRequestBuilder(request);
-  await this.#ensureFunctionAbi(builder);
+  await this.#ensureFunctionAbis(builder);
   const initRequest = builder.build();
   const txExecutionRequest = await this.getTxExecutionRequest(initRequest);
   const provenTx = await this.proveTx(txExecutionRequest, request.executionResult);
@@ -730,7 +756,7 @@ async send(request: UserRequest): Promise<UserRequest> {
     throw new Error('Tx must be proven before sending');
   }
   const builder = new UserRequestBuilder(request);
-  await this.#ensureFunctionAbi(builder);
+  await this.#ensureFunctionAbis(builder);
   const initRequest = builder.build();
   const txExecutionRequest = await this.getTxExecutionRequest();
   const txHash = await this.sendTx(txExecutionRequest, request.tx);
@@ -743,6 +769,8 @@ async send(request: UserRequest): Promise<UserRequest> {
 #### `UserRequest` is a kitchen sink
 
 The `UserRequest` object is a bit of a kitchen sink. It might be better to have a `DeployRequest`, `CallRequest`, etc. that extends `UserRequest`.
+
+Downside here is that the "pipeline" it goes through would be less clear, and components would have to be more aware of the type of request they are dealing with.
 
 #### Just shifting the mutable subclass problem
 
--- a/in-progress/6973-refactor-base-contract-interaction.md
+++ b/in-progress/6973-refactor-base-contract-interaction.md
@@ -320,6 +320,7 @@ export interface SimulationOutput {
   executionResult: ExecutionResult;
   publicOutput: PublicSimulationOutput;
   privateOutput: NestedProcessReturnValues;
+  error?: SimulationError;
 }
 
 export interface UserAPI {
@@ -665,7 +666,7 @@ async simulate(userRequest: UserRequest): {
   // If we're paying, e.g. in bananas, figure out how much AZT that is.
   // Note: paymentMethod.getEquivalentAztBalance() may call `read` internally.
   const equivalentAztBalance = await builder.paymentMethod.getEquivalentAztBalance();
-  gasEstimator = GasEstimator.fromAztBalance(equivalentAztBalance);
+  gasEstimator = new BinarySearchGasEstimator(equivalentAztBalance);
   builder.setGasSettings(gasEstimator.proposeGasSettings());
 
   while (!gasEstimator.isConverged()) {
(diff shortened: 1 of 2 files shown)


iAmMichaelConnor (commented):
Great doc! I've only skimmed it, since the last time I interacted with contracts I was using Truffle, so others will have more modern opinions.
Some Q's:
- At what stage does the PXE get populated with data about the tx that the user has submitted? Is it `simulate`, `prove`, or `send`?
    - Is some data stored at the time of submission? And some at finalisation? And some when decrypting outgoing logs (although I suppose outgoing logs only need to be processed if re-syncing from scratch?)?
- With the wip block building design, there will be several stages to finalising a tx. This is just something to be aware of when designing the promise that's returned by `send()`.
        - Arrival in the tx pool
        - Tx included in a block by the sequencer (a "preconfirmation") on L2 p2p network.
        - Block (in which the tx was included) is submitted to L1, with validator votes (or a commitment bond) but not-yet proven.
            - This submission could be reverted if there's an L1 reorg.
            - This submission could also be reverted if the 51% of signatures / the bond are lying.
        - Block is eventually proven and L1 verifies it as correct.
            - This proof verification could be reverted if there's an L1 reorg, so proper finalisation would need to wait sufficient etc epochs for proper eth finalisation.
- If a user or an app wants to store a record of all inputs to every nested function call of a tx (all args, oracle calls and responses, capsule data, (reverse capsule data?)), so that the tx can be replayed exactly in future, how can that be achieved? Is sufficient data being returned by `simulate` / `prove` / `send`?
- At the moment, I think we're missing auth steps (because we haven't got round to thinking about them). Steps such as (pasted from a list elsewhere):
    - The PXE should prevent apps from accessing other apps' secrets via oracle calls
    - The PXE should check whether the code about to be run actually does belong to a particular contract address
    - The PXE might need to call back to a user/wallet/dapp to ask "Hey, this contract address needs this capsule data to continue - provide it at your own risk"
    - The PXE might need to convey to a dapp "Hey, I've simulated this tx and here are the details, here's what's going to be made public, and it's going to cost this much"
    - The dapp says to the wallet "Please sign this mysterious message".
    - The dapp says to the wallet "Please sign this mysterious transaction"
    - Can this interface support these use cases?
        - How would we achieve callbacks back to the dapp?
- What happens if, during `simulate`, the source code / bytecode for a nested function call isn't available? Can the `simulate` function gracefully pause to download the data, or call back to the dapp to request the data, or somehow get the data?
- What happens if, during `prove`, the proving key / vk for a nested function call isn't available? Can the `prove` function gracefully pause to generate the proving key / vk from the bytecode?

in-progress/6973-refactor-base-contract-interaction.md
```diff
@@ -0,0 +1,1088 @@
...
+// Used by simulate and read
+async #simulateInner(userRequest: UserRequest): ReturnType<BaseWallet['simulate']> {
+  const txExecutionRequest = await this.getTxExecutionRequest(userRequest);
+  const simulatedTx = await this.simulateTx(txExecutionRequest, builder.simulatePublicFunctions, builder.from); 
+  const decodedReturn = decodeSimulatedTx(simulatedTx, builder.functionAbi);
+  return {
+    tx: simulatedTx.tx,
+    publicOutput: simulatedTx.publicOutput,
+    privateOutput: simulatedTx.privateReturnValues,
+    executionResult: simulatedTx.executionResult,
```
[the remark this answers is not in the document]
just-mitch: `tx` is our transaction object (PrivateKernelTailCircuitPublicInputs, private proof, logs, enqueued public functions)

`publicOutput` is 
```ts
/**
 * Outputs of processing the public component of a transaction.
 */
export class PublicSimulationOutput {
  constructor(
    public encryptedLogs: EncryptedTxL2Logs,
    public unencryptedLogs: UnencryptedTxL2Logs,
    public revertReason: SimulationError | undefined,
    public constants: CombinedConstantData,
    public end: CombinedAccumulatedData,
    public publicReturnValues: NestedProcessReturnValues[],
    public gasUsed: Partial<Record<PublicKernelType, Gas>>,
  ) {}
```

The `privateReturnValues` are:
```ts
/** Return values of simulating a circuit. */
export type ProcessReturnValues = Fr[] | undefined;

/** Return values of simulating complete callstack. */
export class NestedProcessReturnValues {
  values: ProcessReturnValues;
  nested: NestedProcessReturnValues[];
```

and `executionResult` is a bit redundant in that it includes all the stuff in `tx`, but also stuff we need for proving like the verification key. It may be that that is all we truly need from this.

just-mitch:
Thank you for the pass @iAmMichaelConnor ! Great questions.

> At what stage does the PXE get populated with data about the tx that the user has submitted?

The transient auth witnesses, capsules, and contracts should be populated in the PXE during *every* call (simulate/prove/send), but *always* removed when the call finishes. This should be an in-memory store that gets initialized for each request and acts as a cache over the persistent lmdb store the pxe has, which is what we do for authwits today.

Regarding persistent storage, I think the PXE should have a configuration parameter that specifies the minimum stage in finalization reached by blocks it receives from its node, e.g. `"submitted" | "proven" | "finalized"`. 

A user will need to configure their PXE to something other than "finalized" if they want to build ahead, in which case we need a way to overwrite the persistent store.

Regarding `.wait`, I think we should add an option such as `until` which accepts one of those "block finalization stages", possibly unioned with `"tx-pool" | "included"`

> replayed exactly

Yes the `TxExecutionRequest` returned by `getTxExecutionRequest` should allow any pxe/node with the required acir/bytecode to run the transaction.

> What happens if, during `simulate`, the source code / bytecode for a nested function call isn't available?

Presently the PXE/node will error, and this work is independent from that. I don't see a reason why, in the future, a PXE couldn't inspect a `TxExecutionRequest` and try to download the contract from a source, then verify that the artifact hash matches what the user expects to run. Similarly, the node could check that a contract class has been registered and ask a peer for the requisite bytecode and verify it by recomputing the contract class ID.

> gracefully pause to generate the proving key / vk

I think this too is independent from this interface, but yes, I don't see why we couldn't, similar to the above. 

> auth steps

Interesting! We could add some kind of oracle "trace/audit" to the simulation output to show the user the access requests of everything.

So it wouldn't be like an interactive "pause mid-simulation and ask the user for permission", it would just be a more complete "I ran the simulation and here is who asked for what". The wallet could display that information to the user after simulation but before proving/submitting, similar to the gas costs.

I think that would cover most of the points there. I believe the PXE already can/should check that the code about to be run belongs to the specified instance using information in the contract class.

in-progress/6973-refactor-base-contract-interaction.md
```diff
@@ -0,0 +1,1092 @@
...
+  functionAbi?: FunctionAbi;
+}
+
+// new
+export interface UserRequest {
+  calls: UserFunctionCall[];
+  gasSettings?: GasSettings;
+  paymentMethod?: FeePaymentMethod;
+  from?: AztecAddress;
+  simulatePublicFunctions?: boolean;
```
LHerskind, line 301: Quick thought. Is it actually useful to have this option? In my mind, you either simulate the full thing and not at all. I get we need some simulation for proving still, but I guess I would remove the option entirely and then have them rely on using `simulate` to simulate instead of the proving as well 🤷. So if you wanna know you do simulate + prove, but if you just try to blast as quickly as possible you will just do prove.

in-progress/6973-refactor-base-contract-interaction.md
```diff
@@ -0,0 +1,1092 @@
...
+  adaptTxExecutionRequest: TxExecutionRequestAdapter;
+}
+
+export interface FeePaymentMethod extends TxExecutionRequestComponent {
+  getEquivalentAztBalance(): Promise<Fr>;
+}
+
+export interface SimulationOutput {
+  tx: Tx;
+  result: DecodedReturn | [];
```
LHerskind, line 318: This one seem slightly odd to me, you mean `DecodedReturn[]`?

in-progress/6973-refactor-base-contract-interaction.md
```diff
@@ -0,0 +1,1092 @@
...
+  executionResult: ExecutionResult;
+  publicOutput: PublicSimulationOutput;
+  privateOutput: NestedProcessReturnValues;
+  error?: SimulationError;
+}
+
+export interface UserAPI {
+  getTxExecutionRequest(userRequest: UserRequest): Promise<TxExecutionRequest>;
+  simulate(userRequest: UserRequest): Promise<SimulationOutput>;
+  read(userRequest: UserRequest): Promise<SimulationOutput>;
```
LHerskind, line 329: Is the `read` essentially just a alias for `simulate` here?

in-progress/6973-refactor-base-contract-interaction.md
```diff
@@ -0,0 +1,1092 @@
...
+export interface AccountInterface extends AuthWitnessProvider, TxExecutionRequestComponent {
+  getCompleteAddress(): CompleteAddress;
+  getAddress(): AztecAddress;
+  getChainId(): Fr;
+  getVersion(): Fr;
+
+}
+
+// unchanged
+export interface AuthWitnessProvider {
```
LHerskind, line 346: As mentioned earlier, I think it is pretty useful to have the Wallet and Account use different cases here, since one have access to information that it can then use to populate the call to the other. e.g., the wallet know chainId and version and just compute the messageHash that is provided to the account.

in-progress/6973-refactor-base-contract-interaction.md
```diff
@@ -0,0 +1,1092 @@
...
+}
+
+```
+
+
+## Implementation
+
+### Expose an explicit `proveTx` on PXE
+
+We currently don't have a way to *only* prove a transaction; everything must be simulated first.
```
LHerskind, line 367: Not sure I agree with the "everything" you could ignore the public part though?

in-progress/6973-refactor-base-contract-interaction.md
```diff
@@ -0,0 +1,1092 @@
...
+
+Then we just have `EntrypointPayload` as a single, concrete class.
+
+### BaseWallet.getTxExecutionRequest
+
+Consider that we have, e.g.:
+
+```ts
+{
+  calls: [{
```
LHerskind, line 386: While this interface might make it more clear, for cases where it is not a multi-call it seems worse to me, seems more tricky to get nice autocomplete for and early compiler error or type errors?

in-progress/6973-refactor-base-contract-interaction.md
```diff
@@ -0,0 +1,1092 @@
...
+    type: functionAbi.functionType,
+    to: instanceAddress,
+    isStatic: functionAbi.isStatic,
+    returnTypes: functionAbi.returnTypes,
+  });
+}
+
+```
+
+#### main function calls
```
LHerskind, line 434: The "main" part is a little weird to me. If you are developing, I would just see it as adding a function call not a main one, what is the mental model difference? To distinguish between inner calls being made? But a user would not be doing those anyway 🤷

in-progress/6973-refactor-base-contract-interaction.md
```diff
@@ -0,0 +1,1092 @@
...
+      .setOrigin(this.address)
+      .setTxContext({
+        chainId: this.chainId,
+        version: this.version,
+        gasSettings: userRequest.gasSettings,
+      })
+      .setFunctionSelector(
+        FunctionSelector.fromNameAndParameters(abi.name, abi.parameters)
+      )
+      .setFirstCallArgsHash(entrypointPackedArgs.hash());
```
LHerskind, line 551: The `firstCallArgsHash` is not super telling to me 🤷 Think it should either be clearer from the name, or do the maker way of something that you don't "assume" what does.

in-progress/6973-refactor-base-contract-interaction.md
```diff
@@ -0,0 +1,1092 @@
...
+async simulate(userRequest: UserRequest): {
+    tx: simulatedTx.tx,
+    publicOutput: simulatedTx.publicOutput,
+    privateOutput: simulatedTx.privateReturnValues,
+    executionResult: simulatedTx.executionResult,
+    result: decodedReturn,
+    request: initRequest,
+} {
+  // If we're simulating, we need to have the payment method set.
+  // Users should use `read` if they just want to see the result.
```
LHerskind, line 653: It is not clear to me that I would be using a "read" if I just want the results when it is not a `view` function. For `viem` I think the `read` is only used for `view` and `pure` functions? It is not uncommon to have something like a simulate a real call that is not static but still to get some return values from it, so seems slightly annoying we we have to deal with a payment method for something like that?

in-progress/6973-refactor-base-contract-interaction.md
```diff
@@ -0,0 +1,1092 @@
...
+      throw new Error(`Expected exactly 0 setup function calls, got ${calls.length}`);
+    }
+    const payload = EntrypointPayload.fromFunctionCalls(builder.appFunctionCalls);
+    const abi = this.getEntrypointAbi();
+
+    const entrypointPackedArgs = PackedValues.fromValues(encodeArguments(abi, [payload, this.completeAddress.address]));
+    const functionSelector = FunctionSelector.fromNameAndParameters(abi.name, abi.parameters);
+
+    const innerHash = computeInnerAuthWitHash([Fr.ZERO, functionSelector.toField(), entrypointPackedArgs.hash]);
+    const outerHash = computeOuterAuthWitHash(
```
LHerskind, line 800: This was updated lately to be a bit more clear 🫡. 
```typescript
const outerHash = computeAuthWitMessageHash(
      { consumer: this.dappEntrypointAddress, innerHash },
      { chainId: new Fr(this.chainId), version: new Fr(this.version) },
);
```

in-progress/6973-refactor-base-contract-interaction.md
```diff
@@ -0,0 +1,1092 @@
...
+      this.dappEntrypointAddress,
+      new Fr(this.chainId),
+      new Fr(this.version),
+      innerHash,
+    );
+    const authWitness = await this.userAuthWitnessProvider.createAuthWit(outerHash);
```
LHerskind, line 806: This is just `this.createAuthWit()`, can be written as that to just make it easier to see that they are the same functions though? Think the `createAuthWit` might just be using a variable that is not defined actually.

in-progress/6973-refactor-base-contract-interaction.md
```diff
@@ -0,0 +1,1092 @@
...
+      .setFunctionSelector(functionSelector)
+      .setFirstCallArgsHash(entrypointPackedArgs.hash)
+      .setArgsOfCalls([...payload.packedArguments, entrypointPackedArgs])
+      .addAuthWitness(authWitness);
+
+  }
+
+
+  createAuthWit(messageHash: Fr): Promise<AuthWitness> {
+    return this.authWitnessProvider.createAuthWit(messageHash);
```
LHerskind, line 824: Here. The `authWitnessProvider` is undefined, it is `userAuthWitnessProvider`.

in-progress/6973-refactor-base-contract-interaction.md
```diff
@@ -0,0 +1,1092 @@
...
+interface SearchRatios {
+  daToL2: number;
+  daTearDown: number;
+  l2TearDown: number;
+}
+
+// Finds a balance that is enough to cover the gas costs of the simulation.
+// Marks as converged if the simulation did not run out of gas,
+// or if it is not possible to increase the balance further.
+export class BinarySearchGasEstimator implements GasEstimator {
```
LHerskind, line 913: Not super into how you are doing much gas estimation, so I will mostly ignore this for now.

in-progress/6973-refactor-base-contract-interaction.md
```diff
@@ -0,0 +1,1092 @@
...
+// Does not modify PXE state.
+const { wallet: aliceWallet, deploymentArgs } = await getSchnorrWallet(pxe, secretKey, salt);
+
+const aliceContractInstance = getContractInstanceFromDeployParams(
+  SchnorrAccountContract.artifact, 
+  deploymentArgs
+);
+
+// Register the account in PXE, and wait for it to sync.
+await aliceWallet.registerAccountInPXE({ sync: true });
```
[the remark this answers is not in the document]
just-mitch: True. This would probably be a two step:
```ts
aliceWallet.registerAccountInPXE()
// later, in some high level context in the UI
aliceWallet.waitForAccountSynch()
// show loading spinner
```

in-progress/6973-refactor-base-contract-interaction.md
```diff
@@ -0,0 +1,1092 @@
...
+  bananaCoinDeploymentArgs
+);
+
+const { request: deployTokenRequest } = await aliceWallet.simulate({
+  calls: [{
+    artifact: TokenContract.artifact,
+    instance: bananaCoinInstance,
+    functionName: bananaCoinDeploymentArgs.constructorName,
+    args: bananaCoinDeploymentArgs.constructorArgs,
+    deploymentOptions: {
```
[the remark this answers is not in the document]
just-mitch: I think it makes sense to keep the constructor as the "main function" (i.e. the one specified by functionName) since it is no different from another function call.

The `deploymentOptions` are meant to provide syntactic sugar for "add ons" that add function calls to our AppPayload for our account contract entrypoint. However, I had had this option before I converted `calls` to take an array, so I can be easily persuaded to drop `deploymentOptions` in favor of just having the explicit calls:

```ts
const tokenContractClass = getContractClassFromArtifact(TokenContract.artifact);
aliceWallet.simulate({
  calls: [{
    artifact: TokenContract.artifact,
    instance: bananaCoinInstance,
    functionName: bananaCoinDeploymentArgs.constructorName,
    args: bananaCoinDeploymentArgs.constructorArgs
  }, {
    ...getCanonicalClassRegisterer(),
    functionName: 'register',
    // would need to add `capsules` to UserFunctionRequest`, but that's probably good.
    capsules: [bufferAsFields(tokenContractClass.packedBytecode)],
    args: {
      artifact_hash: tokenContractClass.artifactHash,
      private_functions_root: tokenContractClass.privateFunctionsRoot,
      public_bytecode_commitment: tokenContractClass.publicBytecodeCommitment
    }
  }, {
    ...getCanonicalInstanceDeployer(),
    functionName: 'deploy',
    args: {
        salt: bananaCoinInstance.salt,
        contract_class_id: bananaCoinInstance.contractClassId,
        initialization_hash: bananaCoinInstance.initializationHash,
        public_keys_hash: bananaCoinInstance.publicKeysHash,
        universal_deploy: bananaCoinInstance.deployer.isZero(),
    }
  }]
})
```

This would also make answering your question below about delayed public deployments easy.

in-progress/6973-refactor-base-contract-interaction.md
```diff
@@ -0,0 +1,1092 @@
...
+Further, we expand `TxExecutionRequest` to allow passing things that we were previously setting directly on the PXE, so simulations can be idempotent.
+
+The interfaces are as follows:
+
+```ts
+
+// new
+export interface ArtifactAndInstance {
+  artifact: ContractArtifact;
+  instance: ContractInstanceWithAddress;
```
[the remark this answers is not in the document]
just-mitch: Yeah I agree. The type is currently:
```ts
export interface ContractInstance {
  /** Version identifier. Initially one, bumped for any changes to the contract instance struct. */
  version: typeof VERSION;
  /** User-generated pseudorandom value for uniqueness. */
  salt: Fr;
  /** Optional deployer address or zero if this was a universal deploy. */
  deployer: AztecAddress;
  /** Identifier of the contract class for this instance. */
  contractClassId: Fr;
  /** Hash of the selector and arguments to the constructor. */
  initializationHash: Fr;
  /** Optional hash of the struct of public keys used for encryption and nullifying by this contract. */
  publicKeysHash: Fr;
}

export type ContractInstanceWithAddress = ContractInstance & { address: AztecAddress };
```

However, we can compute the address from the `ContractInstance`. I'm not sure of the utility of having a type without the address, and would prefer to consolidate if possible. @spalladino ?

in-progress/6973-refactor-base-contract-interaction.md
```diff
@@ -0,0 +1,1092 @@
...
+  ) {}
+  // ...
+}
+
+// new
+export interface InstanceDeploymentParams {
+  constructorName: string;
+  constructorArgs: any;
+  salt: Fr;
+  publicKeysHash: Fr;
```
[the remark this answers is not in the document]
just-mitch: RE the keys instead of the hash: that is a question for @spalladino. 

RE user confusion: I'm hopeful that we can provide objects instead of arrays of arguments to cut down on that, and just have utilities that inspect the function abi to convert the objects into arrays with the proper ordering.

in-progress/6973-refactor-base-contract-interaction.md
```diff
@@ -0,0 +1,1092 @@
...
+  contractArtifact?: ContractArtifact;
+  functionAbi?: FunctionAbi;
+}
+
+// new
+export interface UserRequest {
+  calls: UserFunctionCall[];
+  gasSettings?: GasSettings;
+  paymentMethod?: FeePaymentMethod;
+  from?: AztecAddress;
```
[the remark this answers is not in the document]
just-mitch: Yes, exactly, it is to be used just for simulation. It gets passed through to PXE as `msgSender`:
```ts
  public async simulateTx(
    txRequest: TxExecutionRequest,
    simulatePublic: boolean,
    msgSender: AztecAddress | undefined = undefined,
  ): Promise<SimulatedTx> { ... }
```

in-progress/6973-refactor-base-contract-interaction.md
just-mitch: Yeah I agree. It is in there because it was there already. If no one has strong opinions otherwise, I'd say we nuke it.

in-progress/6973-refactor-base-contract-interaction.md
just-mitch: Actually, it was an effort to be able to return "no value". But I just wasn't thinking: `[]` is a valid `DecodedReturn`, as is `{}`.

in-progress/6973-refactor-base-contract-interaction.md
just-mitch: I had meant that I can't pass a previously simulated Tx to the PXE for proving:

```ts
  public proveTx(txRequest: TxExecutionRequest, simulatePublic: boolean): Promise<Tx> {
    return this.jobQueue.put(async () => {
      const simulatedTx = await this.#simulateAndProve(txRequest, this.proofCreator, undefined);
      if (simulatePublic) {
        simulatedTx.publicOutput = await this.#simulatePublicCalls(simulatedTx.tx);
      }
      return simulatedTx.tx;
    });
  }
```

in-progress/6973-refactor-base-contract-interaction.md
just-mitch: I couldn't come up with a way to support multi-call and single call without accepting an array here or doubling the size of the `UserAPI` to have, e.g. `simulateMultiCall`. 

I'm kind of thinking that the typescript that gets generated for each contract can actually still help us a lot, and provide autocomplete etc, but they will just return these (much simpler) `UserFunctionRequests`. Then the fact that `calls` is an array is maybe not so bad.

E.g.

```ts
{
  calls: [bananaCoinInstance.transfer({ 
      from: aliceAddress,
      to: bobAddress,
      value: privateBalance,
      nonce: 0n
    })]
}
```

My hope with this design is to just some solid low-level primitives that we can build more ergonomics on top of after they're a bit more battle tested. So I don't mind forgoing autocomplete in the immediate term if we think we can extend this (or some other) design to support it.

in-progress/6973-refactor-base-contract-interaction.md
just-mitch: Yeah it's to distinguish between things that would be added by, say, deployment/registration. But if we ditch those options as suggested above then "main" is just the function you're calling from the entrypoint (app payload).

in-progress/6973-refactor-base-contract-interaction.md
just-mitch: Totally agree. I think it should be called `entrypointPackedArgsHash`.

in-progress/6973-refactor-base-contract-interaction.md
```diff
@@ -0,0 +1,1092 @@
...
+const salt = 42;
+
+// Returns the wallet and deployment options. 
+// Does not modify PXE state.
+const { wallet: aliceWallet, deploymentArgs } = await getSchnorrWallet(pxe, secretKey, salt);
+
+const aliceContractInstance = getContractInstanceFromDeployParams(
+  SchnorrAccountContract.artifact, 
+  deploymentArgs
+);
```
spalladino, line 55: I'd rather keep this method "hidden" under a higher-level API. The AccountManager was the initial attempt to do that, but it definitely needs more work.

spalladino: IMO it feels odd to have to repeat the `SchnorrAccountContract.artifact` here, which is known by the `getSchnorrWallet` helper.

in-progress/6973-refactor-base-contract-interaction.md
```diff
@@ -0,0 +1,1092 @@
...
+const receipt = await sentTx.wait({
+  // permanently add the transient contracts to the wallet if the tx is successful.
+  // defaults to true.
+  registerOnSuccess: true
+});
+
+
+```
+
+### Deploy Token
```
spalladino, line 96: I'm fine with all the individual steps outlined here, but since most flows will require all these steps in this order without much change, I'd look into hiding them into something higher-level.

in-progress/6973-refactor-base-contract-interaction.md
just-mitch: You're right that viem's read is only view/pure functions, and I agree there there is some clunkiness here to iron out.

Do you think users should be able to simulate if they have no means of actually paying for what they want to simulate?

in-progress/6973-refactor-base-contract-interaction.md
```diff
@@ -0,0 +1,1092 @@
...
+  return rawReturnValues ? decodeReturnValues(functionAbi.returnTypes, rawReturnValues) : [];
+}
+```
+
+### BaseWallet.read
+
+Like `simulate`, but without the gas estimation.
+
+```ts
+async read(userRequest: UserRequest): DecodedReturn | [] {
```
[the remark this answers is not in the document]
just-mitch: Indeed. We could scope it to only support static calls, or just nuke it in favor of `simulate`.

in-progress/6973-refactor-base-contract-interaction.md
just-mitch: The only difference between the two as the code presently exists (or at least when I wrote this) is that `this.createAuthWit` actually adds the result to the PXE which is not what is desired here.

in-progress/6973-refactor-base-contract-interaction.md
```diff
@@ -0,0 +1,1092 @@
...
+### Major Interface Update
+
+Remove `BaseContractInteraction` and its subclasses.
+
+Define the following:
+
+```ts
+export interface UserAPI {
+  getTxExecutionRequest(userRequest: UserRequest): Promise<TxExecutionRequest>;
+  simulate(userRequest: UserRequest): Promise<SimulationOutput>;
```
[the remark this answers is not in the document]
just-mitch: Yes we can use `from` in the `UserRequest` which PXE will use as `msg_sender`. So I think this use case is covered?

spalladino (commented):
Thanks for taking this on Mitch! I left several comments around the doc, but I think they mostly boil down to:

- Not reusing the same types with optional fields for different things
- Disagree on removing the Contract abstraction
- Agree on reviewing the PXE API
- Agree on having a standard for "tweaking" a tx execution request as it goes through payment methods and entrypoints (though I'm not sold on the builder pattern)
- Unclear on how dapps override wallet entrypoints
- Clearly outlining what goes into a high-level aztec-js, a low-level set of helpers, and a wallet (or UserAPI) interface

If we agree on not killing the Contract class, I'd propose working this from higher to lower level instead, by first cleaning up the ContractInteraction, and from there deriving what are the lower-level helpers and Wallet functions we need. Happy to chat more about this on tomorrow's sync.

in-progress/6973-refactor-base-contract-interaction.md
```diff
@@ -0,0 +1,1092 @@
...
+  read(userRequest: UserRequest): Promise<SimulationOutput>;
+  prove(userRequest: UserRequest): Promise<UserRequest>;
+  send(userRequest: UserRequest): Promise<SentTx>;
+}
+
+export type Wallet = UserAPI & AccountInterface & PXE & AccountKeyRotationInterface;
```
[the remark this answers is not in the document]
just-mitch: I agree. I'm not sure what the separation here should be.

Do you think refactoring that should be part of this design?

in-progress/6973-refactor-base-contract-interaction.md
just-mitch: Not quite. As discussed below, `read` allows user to forgo a fee payment method, but as discussed further below, I think we may drop `read`.

in-progress/6973-refactor-base-contract-interaction.md
just-mitch: So are you thinking something like:
```ts
class Account implements AccountInterface, AccountKeyRotationInterface {
  // ...
}

class Wallet implements UserAPI {
  constructor (public account: Account)
  // ...
}
```

in-progress/6973-refactor-base-contract-interaction.md
```diff
@@ -0,0 +1,1092 @@
...
+Remove `BaseContractInteraction` and its subclasses.
+
+Define the following:
+
+```ts
+export interface UserAPI {
+  getTxExecutionRequest(userRequest: UserRequest): Promise<TxExecutionRequest>;
+  simulate(userRequest: UserRequest): Promise<SimulationOutput>;
+  read(userRequest: UserRequest): Promise<SimulationOutput>;
+  prove(userRequest: UserRequest): Promise<UserRequest>;
```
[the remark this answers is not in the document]
just-mitch: Thinking is that the `UserRequest` will get filled in as you call simulate/prove. E.g. the `proof` is optional unless you're calling `send`. We could certainly create types that enforce properties to be defined if we think that would make things clearer.

in-progress/6973-refactor-base-contract-interaction.md
just-mitch: Totally makes sense! This was a little effort to peel back the onion on the AccountManager, but I agree with you.

in-progress/6973-refactor-base-contract-interaction.md
```diff
@@ -0,0 +1,1092 @@
...
+  publicDeploy?: boolean;
+}
+
+export interface UserFunctionCall {
+  contractInstance: ContractInstanceWithAddress;
+  functionName: string;
+  args: any;
+  deploymentOptions?: DeploymentOptions;
+  contractArtifact?: ContractArtifact;
+  functionAbi?: FunctionAbi;
```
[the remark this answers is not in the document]
just-mitch: `functionName` is what you're calling, `args` are the args to it, `deploymentOptions` is used to "stub in" calls to class registerer and instance deployer. I have a comment above now that advocates for removing `deploymentOptions` in favor of explicit multi-calls to the class registerer and instance deployer.

in-progress/6973-refactor-base-contract-interaction.md
```diff
@@ -0,0 +1,1092 @@
...
+
+// new
+export interface UserRequest {
+  calls: UserFunctionCall[];
+  gasSettings?: GasSettings;
+  paymentMethod?: FeePaymentMethod;
+  from?: AztecAddress;
+  simulatePublicFunctions?: boolean;
+  executionResult?: ExecutionResult; // the raw output of a simulation that can be proven
+  tx?: Tx; // a proven tx to send
```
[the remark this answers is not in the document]
just-mitch: Yep! I call this out in concerns. So I agree: different types are better.

in-progress/6973-refactor-base-contract-interaction.md
```diff
@@ -0,0 +1,1092 @@
...
+  // Note: paymentMethod.getEquivalentAztBalance() may call `read` internally.
+  const equivalentAztBalance = await builder.paymentMethod.getEquivalentAztBalance();
+  gasEstimator = new BinarySearchGasEstimator(equivalentAztBalance);
+  builder.setGasSettings(gasEstimator.proposeGasSettings());
+
+  while (!gasEstimator.isConverged()) {
+    const result = await this.#simulateInner(builder.build());
+    gasEstimator.update(result);
+    builder.setGasSettings(gasEstimator.proposeGasSettings());
+  }
```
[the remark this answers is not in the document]
just-mitch: Good point!

in-progress/6973-refactor-base-contract-interaction.md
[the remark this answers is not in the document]
just-mitch: @spalladino I'm missing where `read` would require the wallet to produce an authwit. User could just use the "signerless" wallet and set `from` on their request?

in-progress/6973-refactor-base-contract-interaction.md
```diff
@@ -0,0 +1,1092 @@
...
+    private userAuthWitnessProvider: AuthWitnessProvider,
+    private dappEntrypointAddress: AztecAddress,
+    private chainId: number = DEFAULT_CHAIN_ID,
+    private version: number = DEFAULT_VERSION,
+  ) {}
+
+  async adaptTxExecutionRequest(
+    builder: TxExecutionRequestBuilder,
+    userRequest: UserRequest
+  ): Promise<void> {
```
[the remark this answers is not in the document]
just-mitch: Great point!

in-progress/6973-refactor-base-contract-interaction.md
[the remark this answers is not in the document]
spalladino: Hmm but doesn't the private kernel init check that msg.sender on the first call of the private stack is zero?

in-progress/6973-refactor-base-contract-interaction.md
```diff
@@ -0,0 +1,1092 @@
...
+  }
+  const nodeInfo = await pxe.getNodeInfo();
+  const entrypoint = new DefaultDappInterface(completeAddress, userAuthWitnessProvider, dappAddress);
+  return new AccountWallet(pxe, entrypoint);
+}
+
+
+const schnorr = new SchnorrAccountContract(signingPrivateKey);
+const authWitProvider = schnorr.getAuthWitnessProvider();
+const aliceDappWrappedWallet = await getDappWallet(pxe, aliceAddress, dappAddress, userAuthWitnessProvider);
```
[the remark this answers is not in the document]
just-mitch: Hm this sounds related to what Lasse was talking about above. Seems like the dapp would need to ask the wallet, "hey give me an `AuthWitnessProvider`", the dapp uses it to build a new wallet, and gives it back to the extension. 

Or the dapp could just keep the wallet it constructed and call `simulate` etc on it.

Or perhaps wallet exposes a `setDappWalletFactory(f: (userAuth: AuthWitnessProvider) => Wallet)`. Then wallet can construct the dapp wallet using its Account interface whenever needed. Not sure how caching works in that case though.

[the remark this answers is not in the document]
spalladino: I'm reluctant to have an API between dapp and wallet that requires them to share "live" js objects. Having a plain JSON API feels safer (and more composable). Maybe what we need is for the wallet to expose a `buildAuthwit` method that gets called from the dapp when needed..? But that will probably be cumbersome to users, as they'll get prompts for authwits without much context.

I think this needs a larger conversation about Wallet and PXE APIs (already mentioned this to @iAmMichaelConnor earlier today).

in-progress/6973-refactor-base-contract-interaction.md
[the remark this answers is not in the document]
LHerskind: I'm fine with it being handled separately to this design 👍

in-progress/6973-refactor-base-contract-interaction.md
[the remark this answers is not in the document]
LHerskind: Not having the options seems fairly bloaty, my main confusion here was around wanting to be more explicit, and then doing things implicit for deployments anyway. In my mind deployments are a bit different from most usual contracts, but I get why having the same flow is useful.

in-progress/6973-refactor-base-contract-interaction.md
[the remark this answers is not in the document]
LHerskind: In the current flow, the `from` is fairly useless in the simulations since it is just applying to the very first call, e.g., the call to the account contract where you don't even use the msg_sender. So we will need to deal with that separately.

in-progress/6973-refactor-base-contract-interaction.md
LHerskind: I have forgotten, with changes on fees and such, at what point do we really require the fees to be paid, if it is only the base rollup stage, then you should be able to simulate it entirely without dealing with the fees?

in-progress/6973-refactor-base-contract-interaction.md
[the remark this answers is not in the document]
LHerskind: Is the "from" value not mostly bogus atm? Since the first call is not what you are actually calling. Otherwise the first authwit at least should be possible to skip with the simulation.

just-mitch:
Stale. Cleanup.

PR #3, opened 2024-06-18
Native Merkle Trees

PhilWindle:


in-progress/0003-native-merkle-trees.md
```diff
@@ -0,0 +1,212 @@
...
+
+For each leaf being inserted:
+
+1. Identify the location of the leaf whose value immediately precedes that being inserted.
+2. Retrieve the sibling path of the preceeding leaf before any modification.
+3. Set the 'next' value and index to point to the leaf being inserted.
+4. Set the 'next' value and index of the leaf being inserted to the leaf previously pointed to by the leaf just updated.
+5. Re-hash the updated leaf and update the leaf with this hash, requiring the tree to be re-hashed up to the root.
+
+Unfortunately, this process is very sequential with minimal opportunity for concurrent hashing. Each sibling path must be taken after having updated the 'low leaf' for the previous insertion. We can achieve a reasonable degree of concurrency here though. We first identify all of the 'low-leaf' values that need updating, then for each we merge steps 2 and 5, making a single pass up the tree, retrieving each node's value before overwriting it. We can then schedule each of these tree traversals as a unit of work to be carried out on a thread pool.
```
iAmMichaelConnor, line 171: > Each sibling path must be taken after having updated the 'low leaf' for the previous insertion

Whilst this is true for the circuit code, is this true for the native code?
I wonder if there's an optimisation that doesn't update each low-leaf sequentially, but instead figures out which low leaves need to change (based on the new values being inserted), and then computes a strategy for hashing up the tree with the fewest hashes (and reads and writes) of the leaves and their paths as possible? The paths of many low-leaves will share common tree nodes, higher up the tree.

Also, in the case where a single low-leaf would be updated by two new leaf insertions, then we would only want to read and update that leaf and its path once.

I would see the process as:

For each leaf being inserted:

In memory:
 1. Identify the location of the low-leaf whose `next_value` "points over" the `value` being inserted.
 3. Set the `next_value` and `next_index` to point to the leaf being inserted.
 4. Set the `next_value` and `next_index` of the leaf being inserted to the leaf previously pointed to by the leaf just updated.

Then, once we have all the pointers (`next_value` and `next_index`) of all low-leaves and new leaves figured out, retrieve their sibling paths and paths with a batch read (or multiple batch read, if a single batch read isn't big enough), noting that many leaves' sibling paths will share common nodes.

Then do the minimum amount of hashing needed: any tree node that needs to be recomputed (i.e. any node on a path of one of the leaves, including the root) should only be recomputed once.

iAmMichaelConnor:
Aside: eventually I'd also like to see a minimal storage tree, for the PXE, which only stores the bare-minimum data to spend the user's notes. Huge storage savings are possible, if storage eventually becomes a bottleneck in the browser https://github.com/iAmMichaelConnor/path-updater

in-progress/0003-native-merkle-trees.md
```diff
@@ -0,0 +1,212 @@
...
+
+### LMDB
+
+LMDB is a high performance key-value database allowing for concurrent read/write access and fully ACID transactions. In particular, it supports up to 126 concurrent read transactions. Write transactions can be performed concurrently with reads but we won't use this. The majority of our World State operations only require read access to persisted data.
+
+There are 3 broad categories of World State operations:
+
+#### Reads
+
+Simply reading data from the trees is performed using either `committed` or `uncommitted` state. Committed state is that which has fully settled and is therefore not going to change over the course of building a block. It can only change upon settlement of a new block. Uncommitted reads will read from the pending state, it is not recommended that uncommitted reads are performed by anyone other than a sequencer/prover.
```
just-mitch, line 137: > settlement of a new block

This "block" is a miniblock/batch right? So it is conceivable that the `committed` state can get wiped in the event of an L2 reorg?

in-progress/0003-native-merkle-trees.md
```diff
@@ -0,0 +1,212 @@
...
+
+This design proposes the creation of a set of multi-threaded merkle tree implementations in C++ using LMDB. It builds upon some previous prototyping to develop concurrent indexed tree insertions.
+
+## Implementation
+
+There are many parts to this design, we will walk through them individiually and discuss the choices made at each stage.
+
+### Overall Architecture
+
+A new C++ binary, World State, will be created that will be started by the node software. It will be configured with the location in which Merkle Tree data should be stored. It will then accept and respond with msgpack-ed messages over one or more streams. The initial implementation will simply used stdio, but this will be absrtacted such that this could be replaced by other stream-based mechanisms.
```
just-mitch, line 27: In the prototype branch the source for the tree lives in barretenberg. Is that where you plan to keep it? This feels like it depends on barretenberg, but is not part of it.

in-progress/0003-native-merkle-trees.md
PhilWindle: I think there are 2 cases. The first being the general case, where for each low leaf updated, we need to retrieve the sibling path before modifications are made. The second case, when a block is simply being synched can be more optimal as you say. In this case, no sibling paths are required and we can identify common nodes as you say.

in-progress/0003-native-merkle-trees.md
PhilWindle: This work only really considers the current model of synching a block once it's settled. Not the upcoming work to introduce batches.

With batches we may need to have an additional layer of state. So e.g.

1. Data becomes 'committed' once it has been fully proven.
2. Data is 'checkpointed' once it is in a batch.
3. Data is 'uncommitted' when a sequencer has inserted it into the tree as part of proposing a batch.

Type 3 can be rolled back to 2. And 2 can be rolled back to 1.

in-progress/0003-native-merkle-trees.md
PhilWindle: As I understand it, bb is structured that everything lives under `barretenberg` but anything not proving system related sits under `crypto`. Most of the work I have done builds on some tree-related code that was already under `crypto/merkle-tree`.

just-mitch (approved):

in-progress/0003-native-merkle-trees.md
spalladino, line 27: Why create it as a separate binary instead of a [nodejs cpp add-on](https://nodejs.org/api/addons.html), so we don't need to manage a separate process? I'd also expect the base cost of calling into an add-on to be cheaper than calling into another process.

PhilWindle: This could maybe work too

alexghr (approved):
Looks great!

PR #4, opened 2024-06-28
doc: Unified bbjs interface proposal

spalladino:
Design for https://github.com/AztecProtocol/aztec3-packages/issues/7228

### DIFF
--- a/in-progress/7228-unified-bbjs-interface.md
+++ b/in-progress/7228-unified-bbjs-interface.md
@@ -225,7 +225,7 @@ And are accessible from a `Prover` object that requires the `circuit` and `flavo
 
 ## Bundling
 
-Since it is now bb.js responsibility to install the native bb package, we'd need an `install` script in the `package.json` that mimics `bbup` behaviour and installs the binary that corresponds to the npm package version and current platform, only that it is downloaded local to the package instead of in the user's home.
+Since it is now bb.js responsibility to install the native bb package, we'd need an `install` script (ie script that is automatically triggered when the package is installed) in the bb.js `package.json` that mimics `bbup` behaviour and installs the binary that corresponds to the npm package version and current platform, only that it is downloaded local to the package instead of in the user's home.
 
 ## Change Set
 
@@ -241,3 +241,7 @@ TBD
 
 - Update [Noir documentation](https://noir-lang.org/docs/dev/getting_started/barretenberg/) on how to install bb and prove and verify circuits.
 - :question: Are there public API docs for bb?
+
+## Rejection Reason
+
+We're shifting our goal to having a common interface for WASM and native via IPC in bb itself, not as a wrapper in bb.js. This changes the design so that the API must be granular and stateful, given native invocation will not happen via CLI.
--- a/in-progress/7228-unified-bbjs-interface.md
+++ b/in-progress/7228-unified-bbjs-interface.md
@@ -1,0 +1,243 @@
+| -------------------- | --------------------------------- |
+| Issue | [Unified bb.js interface that delegates to wasm or native via CLI](https://github.com/AztecProtocol/aztec-packages/issues/7228) |
+| Owners | @spalladino |
+| Approvers | @codygunton @ludamad @PhilWindle @TomAFrench @charlielye |
+
+## Executive Summary
+
+We want bb.js to expose a unified proving and verifying interface that delegates to either wasm or native, depending on the context (web vs node). The updated bb.js should also provide the CLI entrypoint for proving and verifying during Noir development, replacing native bb, and provide a simplified interface similar to that of noir-js-backend-barretenberg.
+
+**The goal of this document is to agree on the common interface.**
+
+## Introduction
+
+Making bb.js the main entrypoint for interacting with bb, either wasm or native, will allow us to:
+
+- Iterate the CLI on bb.js instead of cpp
+- Provide an easier unified programmatic interface to clients
+- Consolidate code currently spread across bb.js, noir-js-backend-bb, and aztec's bb-prover
+- Transparently change how we communicate with native bb by keeping it hidden behind bb.js (we can go from calling the native CLI into calling a long-running process via a nicer binary interface)
+- Simplify the installation process for the user (downloading the native extension would be handled by the bb.js npm package)
+
+## Usage scenarios
+
+We expect to use this API for proving and verifying:
+
+- In a node.js app (like Aztec's prover)
+- In a webapp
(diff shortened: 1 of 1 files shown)


in-progress/7228-unified-bbjs-interface.md
```diff
@@ -0,0 +1,243 @@
...
+
+Native implementation:
+
+- `bb gates`
+
+Wasm implementation:
+
+- `acirGetCircuitSizes`
+
+:question: `bb` exposes a single method to get gates with just a boolean "honk recursion". Does this mean that for `UltraPlonk` we can get circuit sizes setting this flag to false, and for `UltraHonk` and `MegaHonk` we set it to true? How about `ClientIVC`? And for `AVM` does it even make sense to ask about gate count?
```
lucasxia01, line 189: This is a hack I've introduced temporarily while Noir figures out how to correctly modify the verify_proof opcode so it can be used to verify proofs of different proof systems. It's currently set to true for UltraHonk only, so that programs that have verify_proof opcodes will link to the UltraHonk recursive verifier when we prove with UltraHonk.

in-progress/7228-unified-bbjs-interface.md
```diff
@@ -0,0 +1,243 @@
...
+- `UltraHonk`
+- `MegaHonk`
+- `ClientIVC`
+- `AVM`
+
+:question: I collected these values from scanning the bb CLI interface. Are they correct? Should we only target a subset of them? Is "flavor" the correct term? Should we spell it properly as "flavour"?
+
+### Interface
+
+#### `prove(circuit: Circuit, witness: Witness, flavor: Flavor): { proof: Buffer }`
```
fcarreiro, line 72: Note that proving the AVM takes quite different inputs (and works different from) other circuits. Ref `bb_prover.ts` and `bb/execute.ts`; this interface doesn't fit as far as I can see. 

Likely same for verification (we now currently pass things from the prover by prepending to the VK, but we'll eventually have to pass public inputs directly I think).

spalladino:
Going to update the design based on feedback from @charlielye to go with a more stateful and granular API, since we should shoot for having a long-lived native bb sooner rather than later.

in-progress/7228-unified-bbjs-interface.md
```diff
@@ -0,0 +1,247 @@
...
+- `bb contract`
+
+Wasm implementation:
+
+- `srsInitSrs(Crs(1))`
+- `acirNewAcirComposer(0)`
+- `acirLoadVerificationKey`
+- `acirGetSolidityVerifier`
+
+#### `getGateCount(circuit: Circuit, flavor: Flavor): { gates: number }`
```
TomAFrench, line 177: We should maintain the opcode count in the response very least.

The `gates` command also now returns the profiling data for the flamegraph, I'd be happy to push this off into a separate "profile" command or even the nonstandard interface for the time being.

in-progress/7228-unified-bbjs-interface.md
```diff
@@ -0,0 +1,247 @@
...
+Wasm implementation (`AVM` not supported)
+
+- `acirGetCircuitSizes`
+- `commonInitSlabAllocator`
+- `srsInitSrs`
+- `acirNewAcirComposer`
+- `acirInitProvingKey`
+- `acirCreateProof | acirProveUltraHonk | acirProveAndVerifyMegaHonk | acirFoldAndVerifyProgramStack` (there is no `acirProveMegaHonk`, hence the `acirProveAndVerifyMegaHonk`)
+
+#### `proveForRecursion(circuit: Circuit, witness: Witness, flavor: Flavor): { proof: Buffer, proofAsFields: Fields, vk: Buffer, vkAsFields: Fields }`
```
TomAFrench, line 89: Currently Noir embeds in the `circuit` whether it should be proven/verified using a recursion-friendly hash function.

in-progress/7228-unified-bbjs-interface.md
```diff
@@ -0,0 +1,247 @@
...
+Wasm implementation (only `UltraPlonk` supported)
+
+- `acirGetCircuitSizes`
+- `commonInitSlabAllocator(size)`
+- `srsInitSrs(size)`
+- `acirInitProvingKey`
+- `acirGetVerificationKey`
+- `acirSerializeVerificationKeyIntoFields`
+
+#### `verify(proof: Buffer, vk: Buffer, flavor: Flavor): { isValid: boolean }`
```
TomAFrench, line 147: We should also allow verification of recursion-friendly proofs as if I'm incorporating a proof generated from someone else into a proof of my own, I want to verify that the proof is correct before I generate my proof.

TomAFrench (commented):
Some small pieces that jump out at me. Will hold off until the new version for anything more.

in-progress/7228-unified-bbjs-interface.md
spalladino: Good to know. Is this embedded in the bytecode of the circuit, or somewhere else in the artifact?

in-progress/7228-unified-bbjs-interface.md
spalladino: Not sure I'm following: how would the interface for that look like? Isn't it just calling `verify` but with the recursion-friendly proof and a vk specific for it? Or is this related to aggregating one proof into another?

spalladino:
Rejecting since we're shifting our goal to having a common interface for WASM and native via IPC in bb itself, not as a wrapper in bb.js. This changes the design so that the API must be granular and stateful, given native invocation will not happen via CLI. Also, the abstractions are different for Plonk vs Honk, so we may want to rethink them around Honk given Plonk may have its days numbered within bb.

in-progress/7228-unified-bbjs-interface.md
TomAFrench: https://github.com/noir-lang/noir/blob/6936a52bb4de2170b6e4c41afd9cf4327d4fbeb1/acvm-repo/acir/src/circuit/mod.rs#L75

It's currently a boolean flag in the `Circuit` struct so not in the "bytecode" if we class that as just the ACIR opcodes but in the binary blob currently sent to barretenberg.

May move to `Program` in future: https://github.com/noir-lang/noir/issues/5364

in-progress/7228-unified-bbjs-interface.md
TomAFrench: Off the top of my head I'm not sure if the `vk` encodes this information, if so then no change is necessary.

PR #5, opened 2024-07-01
feat: Native Merkle Trees as a native NodeJS module

alexghr:
The document proposes a new interface to the native merkle tree implementation from #3 using Node-API and `bindings` instead of message passing.


in-progress/5040-native-merkle-trees-napi.md
```diff
@@ -0,0 +1,323 @@
...
+| Approvers            | @just-mitch @spalladino |
+| Target Approval Date | 2024-07-05 |
+
+## Executive Summary
+
+This document proposes integrating the [Native Merkle Trees database](https://github.com/AztecProtocol/engineering-designs/blob/f9d1a897303c1481c790cecc4616961e1c183622/in-progress/0003-native-merkle-trees.md) directly into the TypeScript project using a native module written in C++ using [Node-API](https://nodejs.org/docs/latest-v18.x/api/n-api.html) rather than message passing.
+
+## Introduction
+
+The original native Merkle tree spec proposed building a `MerkleTreesDb` native binary in C++. The TypScript code would use message passing over streams to communicate with the database. Messages would be passed over stdin/out which would incur the cost of spawning a process and initializing the C++ code every time an operation is performed. This could be optimized by using long-lived processes but it would mean protecting the interface such that only authorized processes can call the interface.
```
PhilWindle, line 14: I don't think we ever intended to spawn a new process every time. IT was always going to be a long-lived process.

alexghr: I itended to remove this paragraph but missed before pushing the commit :facepalm:

in-progress/5040-native-merkle-trees-napi.md
```diff
@@ -0,0 +1,341 @@
...
+}
+```
+
+![40GB RAM](../images/node40gb.png)
+
+### Error handling
+
+Unhandled exceptions in the C++ code will crash the NodeJS process. Errors must be propagated correctly to the JS side if they can not be handled in C++.
+
+The C++ exceptions flag will be turned on at compile time so exception bubble naturally to the JS side. For async code errors should be returned by rejecting the associated promises.
```
spalladino, line 262: I recall @ludamad mentioned this was an issue when generating WASM. I assume it is not a problem in this context, right?

in-progress/5040-native-merkle-trees-napi.md
```diff
@@ -0,0 +1,341 @@
...
+  bool include_uncomitted = info[2].As<Napi::Boolean>();
+
+  // pointer to helper class for async code (gets cleaned up later), see below
+  auto* tree_op = new bb::world_state::TreeOp(env, deferred, [=]() {
+    bb::crypto::merkle_tree::Signal signal(1);
+    bb::fr leaf(0);
+    auto callback = [&](bb::fr& value) {
+      leaf = value;
+      signal.signal_level(0);
+    };
```
spalladino, line 107: Nit: where is `callback` used?

in-progress/5040-native-merkle-trees-napi.md
```diff
@@ -0,0 +1,341 @@
...
+
+The `Napi` namespace on the C++ contains helper classes to deal with JS primitive values. Strings, numbers, bigints, buffers, arrays, typed arrays and even functions can be freely passed between the two environments.
+
+More complex data structures must be serialized/deserialized. We will msgpack for this as it's already implemented in the C++ code
+
+[`Napi::Value` documentation](https://github.com/nodejs/node-addon-api/blob/cc06369aa4dd29e585600b8b47839c1297df962d/doc/value.md)
+
+### Message passing
+
+Instead of exporting a functions for each operation on the world state, we could instead leverage the existing message passing interface, only instead sending the message across the TS/C++ boundary. This would simplify the module initialization code on the C++ side (only requiring we export a single function) and we'd benefit from easily serializing data types with msgpack.
```
spalladino, line 160: If we implemented this, does it make sense to keep maintaining the message passing interface? Feels like it'd be simpler to deprecate that one, and just have one exported function for each actual tree function, right? No strong feelings here though.

spalladino (approved):
Thanks for looking so thoroughly into this! Looks great to me, but given the amount of cpp in here I don't think I'm the right person for approving. I'd try to get one of @charlielye or @ludamad here if possible.

in-progress/5040-native-merkle-trees-napi.md
fcarreiro, line 1: Nice alternative and investigation!

in-progress/5040-native-merkle-trees-napi.md
alexghr: Fixed :grimacing:

in-progress/5040-native-merkle-trees-napi.md
alexghr: I think it'll come down to the kind of async ops that we'll run inside `AsyncWorker` instances and how we design the code (someone with more C++ experience could chime in here). It might be easier to refactor the generic `tree_op_callback` to return a msgpacked value (currently in this spec the callback must return a `bb::fr`) as then it becomes trivial to send it back to JS as a buffer vs serializing to POJOs in the `OnOk` callback.

just-mitch: If we kept the message passing interface, then wouldn't we not need to worry about keeping types in parity across TS and cpp? And we wouldn't need to deserialize `any` in TS? And minor benefit that other clients of bb (if/when that day should come) that want to use these trees without the nodeapi could more easily integrate them?

in-progress/5040-native-merkle-trees-napi.md
```diff
@@ -0,0 +1,364 @@
...
+    bb::crypto::merkle_tree::Signal signal(1);
+    bb::fr leaf(0);
+    auto callback = [&](bb::fr& value) {
+      leaf = value;
+      signal.signal_level(0);
+    };
+    // for illustration purposes only, actual function call will be different
+    _world_state_svc[tree_id].get_leaf_value(leaf_index, include_uncomitted, callback);
+    signal.wait_for_level(0);
+    return leaf;
```
just-mitch, line 111: Just for understanding, this should `deferred.Resolve(leaf)` right?

just-mitch (approved):
Looks good to me! I'm still a fan of a common message interface, even if this nodeapi wraps it, so I'd weigh the pros/cons there, but generally I like this a lot!

just-mitch:
@alexghr Are we still waiting on people's explicit approval? Or has this entered implementation? Should we merge the design?

PR #7, opened 2024-07-05
doc: Batch proving in Circuits and L1

spalladino:
Related to AztecProtocol/aztec3-packages#7346

### DIFF
--- a/in-progress/7346-batch-proving-circuits-and-l1.md
+++ b/in-progress/7346-batch-proving-circuits-and-l1.md
@@ -1,0 +1,259 @@
+|                      |                                         |
+| -------------------- | --------------------------------------- |
+| Issue                | Batch proving in Circuits and L1        |
+| Owners               | @spalladino                             |
+| Approvers            | @LeilaWang @LHerskind @iAmMichaelConnor |
+| Target Approval Date |                                         |
+
+## Summary
+
+With the separation of sequencers and provers, we now want to submit root rollup proofs that encompass multiple blocks. This requires changes in the rollup circuit topology as well as the L1 Rollup contract.
+
+## Circuits
+
+Let's first review the responsibilities of the current merge and root rollup circuits.
+
+### Merge Rollup
+
+The merge rollup circuit has the following inputs:
+
+```rust
+struct MergeRollupInputs {
+  previous_rollup_data: [{
+    public_inputs: BaseOrMergeRollupPublicInputs
+    nested_proof: NestedProof,
+  }; 2]
+}
+```
+
+The circuit then performs the following checks and computations:
+
+- Recursively verifies the left and right inputs
+- Checks that the tree is greedily filled from left to right
+- Checks that constants from left and right match
+- Checks that the end state from left matches the start state from right (ie they follow from each other)
+- Outputs the start of left and end of right
+- Hashes together or sums up any accumulated fields (tx count, effects hashes, accumulated fees, etc)
+- Propagates constants
+
+And outputs:
+
+```rust
+struct BaseOrMergeRollupPublicInputs {
+  constants: {
+    previous_archive: TreeSnapshot,
+    global_variables: { block_number, timestamp, ... }
+  },
+  start: PartialStateReference,
+  end: PartialStateReference,
+  txs_effects_hash: Field,
+  out_hash: Field,
+  accumulated_fees: Field
+  num_txs: Field,
+}
+```
+
+### Root rollup
+
+The root rollup takes the same inputs as merge rollup, plus fields related to L1-to-L2 messaging, and related to updating the archive tree with the new block root.
+
+```rust
+struct RootRollupInputs {
+    previous_rollup_data : [PreviousRollupData; 2],
+
+    l1_to_l2_roots: RootRollupParityInput,
+    l1_to_l2_messages : [Field; NUMBER_OF_L1_L2_MESSAGES_PER_ROLLUP],
+    l1_to_l2_message_subtree_sibling_path : [Field; L1_TO_L2_MSG_SUBTREE_SIBLING_PATH_LENGTH],
+    start_l1_to_l2_message_tree_snapshot : AppendOnlyTreeSnapshot,
+
+    start_archive_snapshot : AppendOnlyTreeSnapshot,
+    new_archive_sibling_path : [Field; ARCHIVE_HEIGHT],
+}
+```
+
+It performs the same checks as the merge circuit, plus:
+
+- Creates a new L1-to-L2 tree snapshot
+- Creates the new block header, which includes the previous archive tree root
+- Updates the archive tree with the new block hash
+
+```rust
+struct RootRollupPublicInputs {
+    archive: AppendOnlyTreeSnapshot,
+    header: {
+      previous_archive: AppendOnlyTreeSnapshot,
+      content_commitment: ContentCommitment,
+      state: StateReference,
+      global_variables: GlobalVariables,
+      total_fees: Field
+    }
+}
+```
+
+### New rollup structure
+
+We propose changing the current rollup structure introducing two new circuits, a **block root rollup** and a **merge block rollup** circuit. The block root rollup circuit acts exactly the same as today's root rollup, grouping multiple base rollup into a tree via merge rollups until it produces a block. The merge block rollup circuits would then merge multiple blocks into a tree, until it reaches a new root rollup that proves a block range.
+
+The tree levels, from top to bottom, would then be:
+
+- Root
+- Block merge
+- Block root
+- Merge
+- Base
+
+### Block root rollup
+
+The block root rollup circuit is the same as today's root rollup circuit, but with its public inputs tweaked so it matches the public inputs from the block merge rollup as well, in the same way as today the base rollup public inputs are tweaked so they match the ones from the merge rollup.
+
+```rust
+struct BlockRootOrBlockMergePublicInputs {
+  previous_archive: AppendOnlyTreeSnapshot, // Archive tree root immediately before this block
+  new_archive: AppendOnlyTreeSnapshot, // Archive tree root after adding this block
+  previous_block_hash: Field, // Identifier of the previous block
+  end_block_hash: Field, // Identifier of the current block
+  out_hash: Field, // Merkle root of the L2-to-L1 messages in the block
+  start_global_variables: GlobalVariables, // Global variables for this block
+  end_global_variables: GlobalVariables, // Global variables for this block
+  fees: [{ recipient: Address, value: Field }; 32], // Single element equal to global_variables.coinbase and total_fees for the block
+}
+```
+
+### Block merge rollup
+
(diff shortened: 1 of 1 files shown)


in-progress/7346-batch-proving-circuits-and-l1.md
```diff
@@ -0,0 +1,259 @@
...
+```diff
+contract Rollup {
+  bytes32 lastArchiveTreeRoot;
+  uint256 lastBlockTimestamp;
++ bytes32 verifiedArchiveTreeRoot;
++ uint256 verifiedBlockTimestamp;
+}
+```
+
+The `last*` fields are updated every time a new block is uploaded, while the `verified*` ones are updated when a proof is uploaded. In the event of a rollback due to failure on proof submission, the `last*` fields are overwritten with the contents of `verified*`.
```
alexghr, line 215: Maybe we need another set of variables to track in-flight proving, otherwise publishing the 33rd block would overwrite the `last*` variables in the rollup which would fail proof validation.

Probably something to think about for proving coordination and having just a verified/last set of variables is enough.

alexghr (approved):
:fire:

iAmMichaelConnor (commented):
I haven't finished, but here's some early comments. :)
Also, I drew this picture. Would this match your understanding? https://miro.com/app/board/uXjVK4BC8Yg=/ Notice also the blobification circuit, which I think should be loosely considered as part of this proposal. If you don't want to think about actual blobs, then perhaps we could think in terms of sha256-hashes, and at least contemplate what data needs to be hashed and emitted when we reach the 'block root' rollup circuit?
Is there any data above the "block" level (i.e. at the supra-block level) that needs to be submitted to blobs? If so, I'm missing yet another blobification circuit to feed the new 'root rollup' circuit.

(I'll be doing some thinking around this too)

in-progress/7346-batch-proving-circuits-and-l1.md
```diff
@@ -0,0 +1,259 @@
...
+```rust
+struct BlockRootOrBlockMergePublicInputs {
+  previous_archive: AppendOnlyTreeSnapshot, // Archive tree root immediately before this block range
+  new_archive: AppendOnlyTreeSnapshot, // Archive tree root after adding this block range
+  out_hash: Field, // Merkle node of the L2-to-L1 messages merkle roots in the block range
+  previous_block_hash: Field, // Identifier of the previous block before the range
+  end_block_hash: Field, // Identifier of the last block in the range
+  start_global_variables: GlobalVariables, // Global variables for the first block in the range
+  end_global_variables: GlobalVariables, // Global variables for the last block in the range
+  fees: [{ recipient: Address, value: Field }; 32] // Concatenation of all coinbase and fees for the block range
```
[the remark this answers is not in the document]
spalladino: Yep, that's correct!

in-progress/7346-batch-proving-circuits-and-l1.md
spalladino: > otherwise publishing the 33rd block would overwrite the last* variables in the rollup which would fail proof validation

Agree! That's why I suggested moving the "prove that this block root is in the current archive tree" to Solidity instead of the root rollup circuit. The proof would have a "last block proven", and the submitter would include a merkle path from it to the current `lastArchiveTreeRoot`. If that root changes, then the submitter needs to generate a new merkle path instead of rebuilding the entire root proof.

To further mitigate against this race condition, we could keep the last N archive tree roots instead of just the latest.

in-progress/7346-batch-proving-circuits-and-l1.md
```diff
@@ -0,0 +1,259 @@
...
+  end_timestamp: Field,
+  end_block_number: Field,
+  out_hash: Field,
+  fees: [{ recipient: Address, value: Field }; 32]
+}
+```
+
+### Empty block root rollup
+
+Since we no longer submit a proof per block, and thanks to @MirandaWood we now have wonky rollups, we no longer need to fill a block with "empty" txs. This means we can discard the empty private kernel circuit.
```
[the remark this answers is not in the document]
spalladino: That hackmd assumed we needed to create one proof per block, which is no longer the case, so we can replace "empty txs" with "empty blocks" in the superblock (name tbd!) proof.

in-progress/7346-batch-proving-circuits-and-l1.md
```diff
@@ -0,0 +1,259 @@
...
+1. Validate the new block header
+2. Check data availability against the availability oracle
+3. Update the `lastArchiveTreeRoot` and `lastBlockTimestamp` in the contract storage
+4. Consume L1-to-L2 messages from the Inbox
+5. Emit an `L2BlockProcessed` event with the block number
+6. Verify the root rollup proof
+7. Insert L2-to-L1 messages into the Outbox using the `out_hash` and block number
+8. Pay out `total_fees` to `coinbase` in the L1 gas token
+
+The first five items can keep being carried out by the `process` method for each new block that gets submitted. Proof verification, L2-to-L1 messages, and fee payment messages are moved to `submitProof`.
```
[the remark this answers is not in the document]
spalladino: > The process name is unclear to me. Would this be better described as submitBlockProposal?

Depends on what we define as a "block proposal". Naming strikes again! If you mean "a block that is part of the unproven chain", then yeah.

> I think perhaps 2 (which will be calling the blob point eval precompile on some earlier-submitted blob) should be deferred until the proof is being verified.

Blobification is not yet clear to me, but yeah, we can defer that until the proof is submitted and verified if it makes sense. FWIW there's a step missing here that is checking the attestations from the validator committee, but that'll come in later.

> The L1-to-L2 messages shouldn't be removed from the L1 box until the proof has been verified

Good point! Do you know if we'd need to soft-delete them at least, in order to prevent them from being consumed in two unproven blocks in a row? Or is there another mechanism to prevent this?

> Validate the new block header - I thought we didn't want to submit potentially malicious state roots at the time of proposing a block? (and hence shouldn't be submitting new block headers to L1 at this stage)

This proposal is assuming we're still uploading tx effects, but yeah, this may change if/when we switch to tx objects.

spalladino:
@iAmMichaelConnor that picture makes sense, yep! At first glance, blobification makes sense, and yeah, it should happen at the "block root" level.

> Is there any data above the "block" level (i.e. at the supra-block level) that needs to be submitted to blobs? If so, I'm missing yet another blobification circuit to feed the new 'root rollup' circuit.

Not that I can think of. The only things that happen at the root-root level are fee payment and L2-to-L1 messages, neither of which need blobspace I believe.

in-progress/7346-batch-proving-circuits-and-l1.md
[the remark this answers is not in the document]
LeilaWang: > The L1-to-L2 messages shouldn't be removed from the L1 box until the proof has been verified

Was wondering about the same thing! I think we need to restore `toConsume` (but not `isProgress`) in the inbox to the point `verifiedArchiveTreeRoot` was set when a proof fails.

Was also considering deferring `l1_to_l2_roots` to be proved in the final root rollup and consume the root in the inbox from `submiteProof`. But this will result in a bigger delay for users to make use of a l1_to_l2_msg. So maybe not a good idea.

> This proposal is assuming we're still uploading tx effects, but yeah, this may change if/when we switch to tx objects.

What's tx objects?

PhilWindle (commented):
Looks sensible to me. Definitely needs input from the other teams however.

in-progress/7346-batch-proving-circuits-and-l1.md
LHerskind: A fairly simple approach that I think I would take for the first version at least, would be to simple store a "list" of the archive values instead of just the tips. 

This could be as:
```solidity
mapping(uint256 blockNumber => bytes32 archive) public archives;
```

Then you keep track of 2 indexes for how far we have added blocks, and how far we have proven.

This would also allow providing a proof that is not the "next", e.g., to avoid the issue where someone is delaying their proof to limit the time you have to interact. We are looking to add some of this in https://github.com/AztecProtocol/aztec-packages/issues/7614.

in-progress/7346-batch-proving-circuits-and-l1.md
```diff
@@ -0,0 +1,263 @@
...
+3. Consume L1-to-L2 messages from the Inbox
+4. Emit an `L2BlockProcessed` event with the block number
+5. Test data availability against the availability oracle
+6. Verify the root rollup proof
+7. Insert L2-to-L1 messages into the Outbox using the `out_hash` and block number
+8. Pay out `total_fees` to `coinbase` in the L1 gas token
+
+The first four items can keep being carried out by the `process` method for each new block that gets submitted. Proof verification, L2-to-L1 messages, and fee payment messages are moved to `submitProof`. Data availability needs more clarity based the interaction between the blob circuits and the point evaluation precompile, but it may be moved entirely to `submitProof`.
+
+As for consuming L1-to-L2 messages, note that these cannot be fully deleted from the Inbox, since we need to be able to rollback unproven blocks in the event of a missing proof, which requires being able to consume those messages again on the new chain.
```
LHerskind, line 232: L1 to L2 messages don't need to be deleted from the inbox, they are not really deleted there, just progressing what tree we are reading. It still needs to perform some action at the "submitBlock", but we can get this fairly easily https://github.com/AztecProtocol/aztec-packages/issues/7592

in-progress/7346-batch-proving-circuits-and-l1.md
spalladino: @LHerskind sounds good as a first approach, given that L1 gas is free!

in-progress/7346-batch-proving-circuits-and-l1.md
```diff
@@ -0,0 +1,263 @@
...
+- Concatenates and outputs the `fees` from both inputs
+- Outputs `sha256(left.out_hash, right.out_hash)` as its own `out_hash`
+- Outputs `previous_archive`, `start_global_variables`, and `previous_block_hash` from `left`
+- Outputs `new_archive`, `end_global_variables`, and `end_block_hash` from `right`
+
+Note that we say that the global variables in `right` "follow" the ones in `left` if:
+
+- `left.chain_id == right.chain_id`
+- `left.version == right.version`
+- `left.block_number == right.block_number + 1`
```
MirandaWood, line 165:
```suggestion
- `left.block_number + 1 == right.block_number`
```
As discussed :)

just-mitch:
@spalladino @MirandaWood @LeilaWang Is this one being implemented? Shall we merge the PR?

spalladino:
@just-mitch Miranda has already implemented most of it. I was waiting on approval from @LeilaWang and @LHerskind, but I believe we can merge anyway.

LeilaWang:
I'll look at the PR now!

LeilaWang (approved):

just-mitch:
> Oh do you mean this doc pr? 😅

yep!

Sounds good! Thanks all!

PR #8, opened 2024-07-16
Spartan v1

just-mitch:
This design and implementation plan describes version 1 of "Spartan", a public network that will be operational on/before 2024-09-16.


just-mitch:
Fix https://github.com/AztecProtocol/aztec-packages/issues/7482

in-progress/7482-sequencer-prover-test-net.md
```diff
@@ -0,0 +1,409 @@
...
+
+**Confirmation Rules**
+Of a chain, the set of conditions that must be met for a block to be considered confirmed.
+
+
+### SPRTN v1.0.0 Chains
+
+We will explicitly support multiple concurrent "chains":
+
+- "Pending Chain" - Reflects blocks published to L1 with their state diffs, but not yet been proven.
```
spalladino, line 87: Not a fan of the "Pending" name, since it [already has semantics in Ethereum](https://docs.chainstack.com/docs/understanding-pending-block-on-ethereum) which differ from these.

in-progress/7482-sequencer-prover-test-net.md
```diff
@@ -0,0 +1,409 @@
...
+
+After this, the block is confirmed in the Pending Chain.
+
+### The SPRTN v1.0.0 Proven Chain
+
+#### Overview
+
+It is the responsibility of the prover selected at the beginning of an epoch to submit the proof of the epoch.
+
+The proof of epoch i must land in epoch i+1.
```
spalladino, line 176: I wouldn't enforce this for v1. If we did, what should the protocol do if the proof doesn't land on time? We don't handle reorgs nicely yet.

in-progress/7482-sequencer-prover-test-net.md
```diff
@@ -0,0 +1,409 @@
...
+  // Add a validator (who has previously staked) to the staking set.
+  // @dev only admin
+  function addValidator(address validator) external;
+
+  // Remove a currently active validator from the staking set.
+  // @dev only admin
+  function removeValidator(address validator) external;
+
+  // Return the currently active sequencer
+  function getCurrentSequencer() view external returns (address);
```
spalladino, line 252: I'm not sure it should be validator's responsibility to determine who the current sequencer/prover is. Anyway, it's easy to move around once we have more clarity.

in-progress/7482-sequencer-prover-test-net.md
```diff
@@ -0,0 +1,409 @@
...
+
+Much of the current rollup contract will remain unchanged.
+
+```sol
+interface IRollup {
+  // Append a block to the pending chain
+  function processPendingBlock(bytes calldata header, bytes32 archive, bytes calldata signatures) external;
+
+  // Prove a block in the pending chain
+  function processProvenBlock(bytes calldata header, bytes32 archive, bytes calldata proof) external;
```
spalladino, line 278: This seems like we're proving one block at a time. Is that the expectation for v1?

in-progress/7482-sequencer-prover-test-net.md
```diff
@@ -0,0 +1,409 @@
...
+- B's header has been made available on L1
+- B's constituent TxEffects have been made available on L1
+
+After this, the block is confirmed in the Pending Chain.
+
+### The SPRTN v1.0.0 Proven Chain
+
+#### Overview
+
+It is the responsibility of the prover selected at the beginning of an epoch to submit the proof of the epoch.
```
PhilWindle, line 174: Do we want to be this strict for version 1? I thought the plan was to push prover coordination for later. Should we initially allow any to submit the proof to keep the chain moving?

Particularly as we can't currently handle re-orgs.

in-progress/7482-sequencer-prover-test-net.md
just-mitch: Let us know if you have a better option!

in-progress/7482-sequencer-prover-test-net.md
just-mitch: This is related to @spalladino and @LHerskind comments below.

How about we keep this *coordination* (we pick a prover for an epoch), but do not worry about any timeliness (so no re-orgs).

In order for the proven chain to advance, all previous proofs must be posted.

Separately, we keep a function on the rollup contract just for verifying whether a proof is valid (not advancing the proven chain), enabling the "free for all" games to be played on SPRTN, DevNet, or both.

in-progress/7482-sequencer-prover-test-net.md
```diff
@@ -0,0 +1,409 @@
...
+  function currentSlot() view external;
+
+  function currentEpoch() view external;
+}
+```
+
+
+### Prover Coordination
+
+The Prover Node will not need to coordinate with sequencers, and will interface with the L2 P2P network and L1 to generate proof requests (and prove them).
```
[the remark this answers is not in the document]
just-mitch: Updated above:
> - Whenever a block is added to the Proven Chain, rewards will be distributed to the participating proposer and prover.

in-progress/7482-sequencer-prover-test-net.md
```diff
@@ -0,0 +1,409 @@
...
+We will have a set of parameters that can be passed to a deployment script to configure the network.
+
+The parameters will be:
+- Number of sequencers
+- Number of provers
+- Number of PXEs
+- Slot time
+- Epoch length
+- Committee size
+- P2P enabled
```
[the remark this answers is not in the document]
just-mitch: This is "won't do" for this version.

joeandrews (approved):

PR #9, opened 2024-07-17
World State Improvements

PhilWindle:


in-progress/world-state/0004-world-state.md
```diff
@@ -0,0 +1,193 @@
...
+
+
+## Introduction
+
+This design builds on the [Native Merkle Trees](./0003-native-merkle-trees.md) design. The development of Native Merkle Trees provides much higher performance both in terms of the speed of individual operations on the trees and opportunities to execute multiple operations concurrently. The work undertaken until now however only focuses on optimising the existing world state implementation. 
+This has a number of shortfalls.
+
+### Snapshotting
+
+The current snapshot implementation provides 2 options. Either a node only offers `current` state or it offers both `current` state and **all** historical state. The intention being that only full archive nodes would need to offer historical state. This has turned out to be an invalid assumption as most nodes will probably want to provide a short history of snapshotted state leaving archive nodes to provide the full history. We will therefore need to provide the ability to configure a window of blocks for which state is available, up the current latest block number. Currently, it is very hard to prune historical state.
```
spalladino, line 20: > The intention being that only full archive nodes would need to offer historical state. This has turned out to be an invalid assumption as most nodes will probably want to provide a short history of snapshotted state leaving archive nodes to provide the full history.

This is especially relevant when a user is proving their client-side tx and they need to request merkle paths against the state root they're working with. If the node advances to a new block while the user is still simulating their tx, they'd have to throw everything out and start over to use the new root, unless they can request merkle paths for a given recent root.

just-mitch (approved):
Wow clever algorithms and nice writeup!

in-progress/world-state/0004-world-state.md
```diff
@@ -0,0 +1,193 @@
...
+
+## Images
+
+Taking the 3-stage (uncommitted, committed, snapshotted) state architecture outlined above, we can make the concept generic and introduce the term `image` representing a cahe of 'pending' state accumulated through a selection of blocks on top of the `finalized` chain state. The world state module will create an `image` for every tree at every block on the pending chain as it is read from L1. Additionally, images can be requested on demand for block-building activities and later destroyed. Images that extend beyond the previous pending epoch will not duplicate the state of that epoch. Instead they will reference the image at the tip of that epoch. All images will reference the `finalized` store.
```
iAmMichaelConnor, line 70: > as it is read from L1

Perhaps we also want to enable nodes to 'optimistically' create images from not-yet-proposed-to-L1 blocks that they've seen in the mempool. Even if a proposal hasn't been finalised on L1, a user might wish to act on their view of the mempool. E.g. the next sequencer might take a risk and make a head start on preparing their block, or an (advanced) user might wish to optimistically begin generating a tx from the state root of the not-yet-proposed-to-L1 block.

iAmMichaelConnor (commented):
Really nice write up - very clever ideas! This should work nicely! Thanks for sharing. I've only added one small question.

Oh, and this question: I recall in aztec connect we were storing tree stumps (I think that's what they were called?), as a way of reducing the amount of storage. Have we dismissed that as being too complicated?

in-progress/world-state/0004-world-state.md
PhilWindle: Yes, a sequencer who is next in line will most definitely want to do this. If they wait until the prior proposal is confirmed on L1 they will only have a short period to build their own proposal.

just-mitch:
@spalladino are we good to merge this one?

spalladino:
> @spalladino are we good to merge this one?

That's for @PhilWindle to decide!

PR #13, opened 2024-08-09
testnet overview

just-mitch:
Fix AztecProtocol/aztec-packages#7520

### DIFF
--- a/in-progress/7520-testnet-overview.md
+++ b/in-progress/7520-testnet-overview.md
@@ -1,0 +1,391 @@
+# [TestNet Design Overview](https://github.com/AztecProtocol/aztec-packages/issues/7520)
+
+|                      |            |
+| -------------------- | ---------- |
+| Owners               |            |
+| Approvers            |            |
+| Target Approval Date | 2024-08-06 |
+
+This document is a system design overview of what is to be delivered as the TestNet, focusing on networks, L1 interactions, governance, and economics.
+
+A fully functional test network is expected to be delivered by Dec 2, 2024. This network will be a publicly available, but with no guarantees of security or stability.
+
+The deployed network will be referred to as "TestNet".
+
+Thus, in the immediate term, Aztec Labs and/or the Aztec Foundation (once set up) will be coordinating the operation of the following networks:
+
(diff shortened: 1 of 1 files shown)


in-progress/7520-testnet-overview.md
```diff
@@ -0,0 +1,391 @@
...
+- The Proven Chain
+- The Finalized Chain
+
+All three chains will be effectively managed by the Aztec Node and the L1 contracts, and will have different guarantees.
+
+Time will be divided into slots, which will be grouped into epochs.
+
+Each slot will have a proposer, who is responsible for proposing a block of transactions.
+
+Each epoch will have a set of validators, who add economic security to the Pending Chain by providing signatures on proposed blocks; ultimately the Pending Chain is a UX feature that allows users to see their transactions with _some guarantee_ before they are proven.
```
joeandrews, line 141: Can we break out what is being signed on here. i.e The validator is signing over the fact that they have the proof for each transaction, and the proof passes.

We can then add a later training wheel module.

joeandrews: (I should have kept reading)

in-progress/7520-testnet-overview.md
```diff
@@ -0,0 +1,391 @@
...
+- maxFee = maxFeePerL2Gas _ l2GasLimit + maxFeePerDAGas _ daGasLimit
+
+There is an additional pair of parameters to support complex flow such as fee abstraction:
+
+- l2TeardownGasLimit: the maximum amount of L2 gas they are willing to consume for the teardown of the transaction
+- daTeardownGasLimit: the maximum amount of DA gas they are willing to consume for the teardown of the transaction
+
+Both of these values are used to "pre-pay" for the public teardown phase of the transaction.
+
+Each L2 block has a fixed L2 gas limit and a DA gas limit.
```
joeandrews, line 315: Please can we add a note that there will be a mechanism to update / compute the required gas price, or maxFeePerDAGas and maxFeePerL2Gas that runs when a block is proven, either in solidity or inside the circuit. Exact mechanism is TBC.

joeandrews (approved):

PR #14, opened 2024-08-21
A proving job broker implementation

PhilWindle:


in-progress/proving-queue/0005-proving-queue.md
just-mitch, line 1: Can we see the typescript interfaces that the orchestrator and brokers are expected to implement?

Are the orchestrator and broker expected to run on different machines? How do they "register" each other?

in-progress/proving-queue/0005-proving-queue.md
```diff
@@ -0,0 +1,97 @@
...
+
+![Proving Architecture](./proving-arch.png)
+
+### A Simple but Sufficient Prover Implementation
+
+Frankly, there are many ways in which the `Prover` component can be built. It is a common pattern within Web 2 and many technologies exist to help. Ideally however we would like to reduce our dependencies as much as possible, so a solution relying solely on existing dependencies would be desirable. First, let us look at our high level requirements:
+
+1. Proving a single block of 11520 transactions (10 TPS for 32 blocks where each block is 3 ethereum blocks)
+2. 3 public functions per transaction results in the other of 100,000 proof requests.
+3. Depending on timeliness requirements and proof computational complexity, we should expect the number of prover agents to be up to 10,000.
```
spalladino, line 32: It looks scary when put this way

in-progress/proving-queue/0005-proving-queue.md
```diff
@@ -0,0 +1,97 @@
...
+``` TS
+type Metadata = {
+  proofType: PROOF_TYPE; //(BASE_ROLLUP etc...)
+  epochNumber: number;
+  status: // (IN_PROGRESS or COMPLETED)
+}
+
+```
+
+Proving request data, such as input witness and recursive proofs are stored in a directory labelled with the job's id residing on an NFS share/S3 bucket or similar. This is an optimsation, as prover agents can access the data independently, without requiring the broker to transfer large amounts of data to them. If it turns out that this is not required then the proof requests will reside on a disk local to the broker and will be transferred over the network from the broker. Maybe this should be a configurable aspect of the system.
```
spalladino, line 55: Unless we see the amount of data being too large, I'd leave this optimization out, since having a shared NFS may not be an option for decentralized proving marketplaces, and an S3 bucket ties the implementation to AWS. I suspect the amount of data we need to push is not _too_ much.

spalladino: Actually, we can probably count on an S3 bucket anyway. It is only coupled to the prover-node, from the prover-agent's perspective it's just a URL that they use for downloading proving work.

in-progress/proving-queue/0005-proving-queue.md
```diff
@@ -0,0 +1,97 @@
...
+The reason for the start time negotiation around agents is that we need to consider a broker crash/restart.
+
+Upon restart, the broker should:
+
+1. Re-build the in-memory cache from the index.
+2. Start accepting new job requests from agents.
+
+The above means that a job that is already being worked could be given to another agent after a restart. All of the data around current proving jobs is only in memory so is lost upon process termination. By negotiating based on start time, soon after restart the system will revert to how it was without too much duplicated effort. Of course, for small proofs, it is still possible that a proof could be completed twice. This seems an acceptable trade-off.
+
+Finally, the broker periodically checks all current jobs to see if their `Last Update Time` is beyond a certain window. If so, the details of the agent proving the job are removed, so they become available again.
```
spalladino, line 93: I'd say we remove this periodic process, and instead when an agent polls we search for jobs that are either not taken or that their last update time is beyond a threshold.

spalladino (approved):
LGTM! I think it's a safe assumption that we can squeeze everything in memory without issues. Otherwise, we can just abuse LMDB key ordering to act as indices (we are already doing something like that in the [tx pool](https://github.com/AztecProtocol/aztec-packages/blob/bfbc4b2126c0fed038d46575a60311234a1a484a/yarn-project/p2p/src/tx_pool/aztec_kv_tx_pool.ts#L19-L22), but here we'd need to also include info like epoch number in the key).

just-mitch:
Is this one being implemented as presented? Shall we merge?

PR #16, opened 2024-08-27
design: Extensions to publishing txobjects

LHerskind:


in-progress/8025-publishing-tx-objects/design.md
```diff
@@ -48,27 +69,159 @@ To the DA oracle contract, the proposer will publish the following data:
...
+  pass
+
+
+def submit_proof(proof, archive):
+  '''
+  Notice that this only supports sequential proofs as we always feed `self.state.proven_archive` as start state.
+  '''
+  proposal_hashes = []
+  block_number_start = self.state.proven_block_count
+  end_slot = get_last_slot_in_epoch(get_epoch_from_slot(self.proposals[block_number_start].slot))
```
just-mitch, line 132: I think this should be 
```python
get_last_slot_in_epoch(get_epoch_from_slot(self.proposals[block_number_start].slot + 1))
```
assuming we want the last slot in the epoch after the most recently proven epoch.

in-progress/8025-publishing-tx-objects/design.md
```diff
@@ -48,27 +69,159 @@ To the DA oracle contract, the proposer will publish the following data:
...
+
+  for bn in range(block_number_start, block_number_start + EPOCH_LENGTH):
+    proposal_log = self.proposals[bn]
+    if proposal_log.slot <= end_slot:
+      proposal_hashes.append(proposal_log.proposal_hash)
+      block_number = bn
+    else:
+      break
+
+  assert len(proposal_hashes) > 0, 'no proposals'
```
just-mitch, line 144: What do we actually want to do in the case where an epoch has no proposals?

in-progress/8025-publishing-tx-objects/design.md
```diff
@@ -48,27 +69,159 @@ To the DA oracle contract, the proposer will publish the following data:
...
 
-Keeps track of the versioned KZG hashes that were present in a transaction.
+Solution 1 is significantly cheaper as it don't update storage. Furthermore, by requiring that the versioned KZG hashes are part of the header provided, we can show the exact KZG hashes using just the transaction and the commitment to the block hash, see `BlockLog.blockHash`.
 
 #### Rollup Contract Changes
+> There is probably some 1-off from the `count` `block_number` like values.
+
+```python
+struct State:
+  pending_proposal_count: uint64;
```
just-mitch, line 88: Storing the `count` versus `index`/`number`/`height` is really messing with me.

in-progress/8025-publishing-tx-objects/design.md
```diff
@@ -48,27 +69,159 @@ To the DA oracle contract, the proposer will publish the following data:
...
+  assert len(proposal_hashes) > 0, 'no proposals'
+
+  prev_archive = self.proposals[self.state.proven_block_count - 1]
+  assert proof.verify(prev_archive, proposal_hashes, archive)
+
+  self.payout_fees();
+
+  self.state.proven_slot = proposal_log[-1].slot
+  self.state.proven_block_count = block_number + 1
+  self.proposals[block_number] = archive
```
just-mitch, line 153: I think you meant to store this in a different struct?

in-progress/8025-publishing-tx-objects/design.md
```diff
@@ -48,27 +69,159 @@ To the DA oracle contract, the proposer will publish the following data:
...
+      return
+    self.forced_inclusion_tip = cache
+
+def submit_proof(proof, archive):
+  # super.submit_proof(proof, archive)
+  
+  forced_tip = self.forced_inclusions[self.forced_inclusions_tip]
+  assert forced_tip.included_by_slot == 0 or forced_tip.included_by_slot > proposal_hashes[-1].slot, 'force'
+
+def submit_proof_with_force(proof, archive, force_inclusion_proofs):
```
just-mitch, line 215: When/where/how does this get called?

in-progress/8025-publishing-tx-objects/design.md
```diff
@@ -48,27 +69,159 @@ To the DA oracle contract, the proposer will publish the following data:
...
+      block_number = bn
+    else:
+      break
+
+  assert len(proposal_hashes) > 0, 'no proposals'
+
+  prev_archive = self.proposals[self.state.proven_block_count - 1]
+  assert proof.verify(prev_archive, proposal_hashes, archive)
+
+  self.payout_fees();
```
just-mitch, line 149: Currently, `total_fees` is in the `Header`, which is supplied to `submitBlockRootProof`. Seems it'll need a way to either pass a list of these `total_fees` to this function and distribute rewards for each block, or aggregate the fees to a final sum for the epoch and pass that into this function and divide the payout evenly.

e.g. if an epoch has 2 proven blocks with fees 10 and 20, the contract can either:
**Option A:** Pass the list [10, 20] and pay out, for example
TX 1: 10 -> 70% to the proposer, 20% to validators, 10% to treasury
TX 2: 20 -> same

**Option B:** Sum [10, 20] in the rollup circuits and pay out 
30 -> 70% to proposers (divided evenly, if you were proposer multiple times you get counted multiple times), 20% to validators (same), 10% to treasury

Speaking of, where/how is the contract going to get the proposer's eth address?

just-mitch: BTW, my intuition tells me option A will be prohibitively expensive.

LHerskind: I think that is essentially what some of Miranda's work does in submit. I just did not bother to have it in here for brevity. Should probably just extend it even if not super concise so there is less confusion.

Providing the data is not necessarily prohibitively expensive, but performing payouts for a full epoch might be, so there is the potential that it is made as a Merkle drop where the proposer needs to claim.

You have the proposer address from the proposer_sig in each proposal using ecrecover

in-progress/8025-publishing-tx-objects/design.md
LHerskind: Will add a comment, but you can see that if a proof is made that includes things that must be forced, you would use it to showcase that it was included, otherwise it might fail due checks at the end

in-progress/8025-publishing-tx-objects/design.md
LHerskind: Will look at trying it to indexes only to not mess with you 👍

in-progress/8025-publishing-tx-objects/design.md
LHerskind: It would just skip the empty epoch. 
It would only be hit if there are literally no proposals. Note that it is just taking all the proposals within the next epoch available to it, not requiring that the next epoch must be +1 of old.

in-progress/8025-publishing-tx-objects/design.md
LHerskind: The index/count is messing with you. Will try to fix them for less confusion

in-progress/8025-publishing-tx-objects/design.md
just-mitch: So the proposer needs to make the determination of which function to hit in the rollup contract?

LHerskind: No, the proof submitter.

in-progress/8025-publishing-tx-objects/design.md
```diff
@@ -48,27 +69,280 @@ To the DA oracle contract, the proposer will publish the following data:
...
+  slot_number: uint256
+  timestamp: uint256
+  coinbase: address
+  fee_recipient: AztecAddress # discuss
+  gas_settings: GasSettings
+
+
+struct Attestations:
+  signature: AggregateBLS
+  missing: uint256 # support up to 3*256 members in committee
```
LHerskind, line 118: Oh, this is only if we can change the ordering such that we have the "missing" at the start. Which likely will cause of a lit more pain, so it is likely much simpler to just store commitee_size / 256 uints for bitmap flags instead.

PR #21, opened 2024-09-05
design: blob background

LHerskind:
Adds a small initial background for blobs. To be completed

### DIFF
--- a/in-progress/8403-blobbity-boo.md
+++ b/in-progress/8403-blobbity-boo.md
@@ -1,0 +1,130 @@
+
+# Template
+
+|                      |                                            |
+| -------------------- | ------------------------------------------ |
+| Issue                | [title](github.com/link/to/issue)          |
+| Owners               | @LHerskind @MirandaWood                    |
+| Approvers            | @just-mitch @PhilWindle @iAmMichaelConnor  |
+| Target Approval Date | YYYY-MM-DD                                 |
+
+
+## Executive Summary
+
+We will retire calldata for block bodies and instead use EIP-4844 blobs.
+
+## Introduction
+
+> Briefly describe the problem the work solves, and for whom. Include any relevant background information and the goals (and non-goals) of this implementation.
+
+### Background
+
+#### Our system 101
+In our current system, we have an `AvailabilityOracle` contract.
+This contract is relatively simple, given a bunch of transactions it will compute a commitment to them, by building a merkle tree from their transaction effects.
+The root of this tree is then stored in the contract such that we can later check for availability.
+
+When a block is proposed to the rollup contract, it will perform a query to the `AvailabilityOracle` to check availability of the `txs_effects_hash` of the header.
+
+Since the only way it could be marked as available was by it being hashed at the oracle, we are sure that the data was published.
+
+When the proof is to be verified, the `txs_effects_hash` is provided as a public input.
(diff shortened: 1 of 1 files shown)


in-progress/8403-blobbity-boo.md
```diff
@@ -0,0 +1,348 @@
...
+
+To evaluate $p$ in evalulation form at some value not in its domain (i.e. not one of the $\omega^i$ s), we use the [barycentric formula](https://dankradfeist.de/ethereum/2021/06/18/pcs-multiproofs.html#evaluating-a-polynomial-in-evaluation-form-on-a-point-outside-the-domain):
+
+$$p(z) = A(z)\sum_{i=0}^{4095} \frac{d_i}{A'(\omega^i)} \frac{1}{z - \omega^i}.$$
+
+What's $A(x)$, you ask? Doesn't matter! One of the nice properties we get by defining $p$ as an interpolation over the roots of unity, is that the above formula is simplified to:
+
+$$p(z) = \frac{z^{4096} - 1}{4096} \sum_{i=0}^{4095} \frac{d_i\omega^i}{z - \omega^i}.$$
+
+We can precompute all the $\omega^i$, $-\omega^i$ s and $4096^{-1}$, the $d_i$ s are our tx effects, and $z$ is the challenge point (discussed more below). This means computing $p(z)$ is threoretically 4096 wrong-field multiplications and 4096 wrong-field divisions, far fewer than would be required for BLS12-381 elliptic curve operations.
```
just-mitch, line 151: Under the current design, we wouldn't actually be computing `p(z)` in a circuit right?

But could/should we?

I.e. is it possible to effectively perform `verify_kzg_proof(commitment, z, y, proof)` in a circuit, and then have the user provide `commitment`, `y`, and `proof`. 

This convinces the circuit that any blob with that commitment has all the tx effects.

Then on L1 we:
1. store the commitments of the blobs at the time the block is proposed
2. match the stored commitments against the user provided ones when circuit verified the opening

That convinces L1 that the blobs the circuit proved against matched the ones that were previously published.

iAmMichaelConnor: @MirandaWood mentioned this as a possible approach to me recently. But she has dismissed it (and I agree) that computing a BLS12-381 pairing inside a Noir circuit would be many many constraints and very difficult to audit.

iAmMichaelConnor: > Under the current design, we wouldn't actually be computing p(z) in a circuit right

The barycentric formula is a way of computing p(z), so technically we _are_ computing it in the circuit with this proposal

just-mitch:
Hey @LHerskind , @iAmMichaelConnor , @MirandaWood , 

Just to confirm, the design presented here is the one we're planning to implement in the immediate term right?

MirandaWood:
> Hey @LHerskind , @iAmMichaelConnor , @MirandaWood ,
> 
> Just to confirm, the design presented here is the one we're planning to implement in the immediate term right?

I do need to make some small changes/clarifications to the write up, but for the maths/rollup circuits sections, generally yes. As for L1, I think @iAmMichaelConnor is still thinking about the multi-commitment multi-opening method so unsure on that.

iAmMichaelConnor:
So this is an old PR, and of course this all got implemented last year. What do we do with this PR?

PR #22, opened 2024-09-05
Proof Timeliness on L1

just-mitch:
Design for https://github.com/AztecProtocol/aztec-packages/issues/8401


LHerskind (commented):
- The amount of arrows can make it slightly difficult to figure follow the diagram easily. Think its fine for now though.
- The first happy path seems slightly, weird, because what I would expect as happy, is that you match the slot that was in the first drawing, but it is not necessarily that because of the other slots 🤷‍♂ 
- Would make it clear that the proof and proposal don't need to be the same person, e.g., in slot `43*E+C+P` you have a proposal and proof, want to make it clear that the proof could come earlier and is independent from the proposal.
- Practically, when the re-org happen, the next proposer have the option to post a fresh block on the pending chain (which is now just the proven because proved).
- `C` have to be large, then the time to prove could be tiny. Also, the first `C` members essentially have the option to censor the entire last epoch + the blocks before it, so we need the change for them all to be "bad" to be tiny. In my mind, it is a hypergeometric distribution, where `n = m = C` and `N` is the committee size.
- Following above, we still have to ensure that `C + P <= E` since it could otherwise give a really weird case where it essentially becomes an option for the second monopoly phase to re-org the first. 
- It is slightly painful, so if you have a large committee `N`, then you need to increase `C` to get a proper assurance that someone will post. Here it can get a little weird, because `E` would ideally be independent of `N`, but now we end up limiting it directly. 
- When we say it grants you monopoly rights, what does that REALLY mean. That you are the only that can submit, or that you are the only that will get rewarded?
	- If only to submit, then there is an issue as you could put down the bond, and then just do nothing. It will essentially end up making a re-org of depth 2 epochs 😬 To get around that it seems like you would need to make the bond fairly large.
- Implementation: Took a quick look, not diving too deep into it
	- We gotta talk about your accounting and separation of concerns. If you inject all the bonding and funds handling into the rollup contract I am going to hunt you down. 
	- I can use other peoples funds for the bond and grief them by not posting proofs.
	- What is `currentProofClaimer` it is a value but not set anywhere.
	- A bunch of variables are never defined before used, or missing types etc.
	- Why do you even store `slotOfClaim`, in my mind you would have until the end of the epoch e.g., up till `C+P` to prove it, not just `P`, you would likely only be using the time after your proposal, but I do not see the gain there, it seems nicer to have that people earlier essentially have more time so you further given them a reason to do it.
		- If you have multiple blocks in a row it would make sense for you to delay the bond to the last of yours since that gives you more time to prove 🤷‍♂ seem counterproductive to what we want.
	- There is a lot of re-computation instead of caching. 
	- The `proposalLog.hash == 0 and != 0` should not exists, it is essentially a check that you are "beyond" the chain, so could be handled just with the chain tips and length. 
	- The `provenBlockNumber` is never assigned.
	- Jumping forword for now

in-progress/8401-proof-timeliness/proof-timeliness.ipynb
```diff
@@ -0,0 +1,164 @@
...
+    "\n"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def calculate_first_p_malicious(V, C, P):\n",
```
LHerskind, line 43: For clarity down here I think it would have been neater using pmf a bit more instead of doing it manually. Only a readability thing really. 🫡 

```python
import numpy as np
from scipy.stats import hypergeom
# Problem statement 
# Draw a committee of 48 validators from a large set of 1000 validators, with 333 malicious validators
# We model this as a hypergeometric distribution (without replacement)
# Then from the smaller committee, compute the probability of drawing k 

N_VALIDATORS = 10000
N_MALICIOUS = N_VALIDATORS // 3
N_COMMITTEE = 48

k = 13

# Step 1: Probability distribution of number of malicious validators in the committee
# We'll calculate the probability of getting `mal_commitee` malicious validators in the committee
prob_malicious_in_committee = np.array(
    [hypergeom.pmf(mal_commitee, N_VALIDATORS, N_MALICIOUS, N_COMMITTEE) for mal_commitee in range(0, N_COMMITTEE+1)]
)

# Step 2: For each possible number of malicious validators in the committee, calculate the probability of drawing 
# malicious malicious validators
prob_drawing_all_malicious = np.zeros_like(prob_malicious_in_committee)
for mal_commitee in range(k, N_COMMITTEE+1):  # We need at least `k` malicious validators to have a non-zero probability
    prob_drawing_all_malicious[mal_commitee] = hypergeom.pmf(k, N_COMMITTEE, mal_commitee, k)

# Total probability: Sum the probabilities of each scenario
total_prob = np.sum(prob_malicious_in_committee * prob_drawing_all_malicious)

print(f"The probability of drawing {k} malicious validators in the first {k} draws from the committee is: {total_prob}")
```

in-progress/8401-proof-timeliness/design.md
```diff
@@ -0,0 +1,268 @@
...
+        if (proofClaim.rewardRecipient != address(0)) {
+            revert Rollup__ProofRightAlreadyClaimed();
+        }
+
+        if (currentSlot % EPOCH_DURATION >= CLAIM_DURATION) {
+            revert Rollup__NotInClaimPhase();
+        }
+
+        // Bond provider must have approved the contract to transfer the bond amount
+        testToken.transferFrom(_bondProvider, address(this), PROOF_COMMITMENT_BOND_AMOUNT);
```
LHerskind, line 148: As earlier. This is vulnerable. But for conveying sentiment it might be fine, just REALLY don't want anyone to actually do this.

in-progress/8401-proof-timeliness/design.md
```diff
@@ -0,0 +1,268 @@
...
+        uint256 currentEpochStartSlot = currentEpoch * EPOCH_DURATION;
+
+        if (proofClaim.rewardRecipient == address(0) &&
+            currentSlot >= currentEpochStartSlot + CLAIM_DURATION) {
+            state.pendingTip = state.provenTip;
+            emit PendingChainPruned(state.pendingTip.blockNumber, state.pendingTip.slotNumber);
+        } else if (proofClaim.rewardRecipient != address(0) &&
+                   currentSlot >= currentEpochStartSlot + EPOCH_DURATION) {
+            state.pendingTip = state.provenTip;
+            _slashBond();
```
LHerskind, line 171: I don't think you actually need a separate function to "slash" bond. As long as it "gets lost" by it being pruned. Think accounting can be done somewhat smart such that we don't have to actually update storage here. Which also means that the two cases become just one which seems to make it slightly simpler to read.

in-progress/8401-proof-timeliness/design.md
```diff
@@ -0,0 +1,268 @@
...
+Fill in bullets for each area that will be affected by this change.
+
+- [ ] Cryptography
+- [ ] Noir
+- [ ] Aztec.js
+- [ ] PXE
+- [ ] Aztec.nr
+- [ ] Enshrined L2 Contracts
+- [ ] Private Kernel Circuits
+- [ ] Sequencer
```
LHerskind, line 235: I know it is saying on L1, but it feels a bit fun to have it on L1, but nothing that is using L1 knows about it 🤷‍♂, e.g., someone needs to know to put up a bond for example, and if we want to use this we would need to also support that part.

LHerskind (approved):
THe implementation outline should be fixed to not be vulnerable, but I can deal with that when we get around to implementation.

in-progress/8401-proof-timeliness/design.md
just-mitch: Agreed. The sequencer changes will be a separate design.

in-progress/8401-proof-timeliness/design.md
just-mitch: Fixed! Flow indicates now that claim proof is to be performed by the bond provider.

in-progress/8401-proof-timeliness/design.md
```diff
@@ -0,0 +1,289 @@
...
+if there are block rewards in test token TST, they will be sent to the address that submitted the claim.
+
+If no claim is submitted, the pending chain is pruned to the tip of the proven chain;
+a caveat is if a proof for epoch `n-1` is submitted during the proof claim phase.
+
+### Proof production
+
+The proof for epoch `n-1` must land in epoch `n`.
+
+Failure to submit a proof results in the bond posted with the claim being slashed if it exists.
```
spalladino, line 49: I think we need a "proof race" or some other method if the proof fails to arrive. Otherwise, the economic cost of rolling back a significant portion of the chain is "just" the prover bond, and we go back to the old "option" issue we had with Fernet before introducing the validator set: a prover (or a proposer, if they submit the prover bond themselves) can decide whether to roll back about 20 minutes of the chain at only the cost of the bond.

spalladino: That said, this is probably good enough for testnet.

just-mitch: I think it is a nice result to force proofs of epoch `n-1` landing in `n` (instead of possibly in `n+1`), so to have an option for a proof race while keeping this property we would likely need to extend the length of the epochs and/or the duration of an L2 slot.

For example, it is possible to have a proof race period if we bump the number of slots in an epoch to 48, and the duration of a slot is 36 seconds, that puts us at 21 minutes after the proof claim period (i.e. enough time for the first proof to not land, and have enough time to make another), and total epoch length of ~29 minutes.

But I agree with you. This design does not preclude a subsequent race, we just need to make sure epochs are timed long enough and then extend the rollup logic.

PR #23, opened 2024-09-06
design: Initial based fallback description

LHerskind:
An initial outline for how a based fallback design could look.


aminsammara:
Sounds good. I have a comment/question on relation to forced inclusion mechanism. 

> Relation to Forced Inclusions: If the L1 contract enforce that there is a minimum block size, the committees only option to try and censor will be to entirely stall the rollup.

1. Even if L1 enforced a min block size, a censoring committee can just send a sufficient amount of cheap txs to avoid "stalling the chain" and thereby delay fallback mode. 

2. I see a similar attack by the committee during fallback mode. They can send a number of dummy txs (that they already hold proofs for) to the forced inclusion list and then submit proof in fallback mode faster than everyone else. They could do this to prevent someone posting a block containing the censored transactions. 

What do you think about: 

1. A certain number of valid txs in the forced inclusion list that stay censored > $t$ time units causes fallback mode to activate, regardless of whether $T_{\textsf{fallback, enter}}$ passed or not. 

2. Once fallback mode has been activated due to censored transactions, the first few blocks accepted must include a certain % of the tx on the forced inclusion list (as of time just prior to fallback mode) to get accepted.

LHerskind:
> Sounds good. I have a comment/question on relation to forced inclusion mechanism.
> 
> > Relation to Forced Inclusions: If the L1 contract enforce that there is a minimum block size, the committees only option to try and censor will be to entirely stall the rollup.
> 
> 1. Even if L1 enforced a min block size, a censoring committee can just send a sufficient amount of cheap txs to avoid "stalling the chain" and thereby delay fallback mode.
> 2. I see a similar attack by the committee during fallback mode. They can send a number of dummy txs (that they already hold proofs for) to the forced inclusion list and then submit proof in fallback mode faster than everyone else. They could do this to prevent someone posting a block containing the censored transactions.
> 
> What do you think about:
> 
> 1. A certain number of valid txs in the forced inclusion list that stay `censored > t` time units causes fallback mode to activate, regardless of whether $T_{\textsf{fallback}, \textsf{enter}}$ passed or not.
> 2. Once fallback mode has been activated due to censored transactions, the first few blocks submitted must include a certain % of the tx on the forced inclusion list (as of time just prior to fallback mode) to get accepted.

First, as all of this is really related to force inclusion, I will just briefly answer here and then pop a link in here when I have written the other doc for forced inclusion.

> 1. Even if L1 enforced a min block size, a censoring committee can just send a sufficient amount of cheap txs to avoid "stalling the chain" and thereby delay fallback mode.
> 2. I see a similar attack by the committee during fallback mode. They can send a number of dummy txs (that they already hold proofs for) to the forced inclusion list and then submit proof in fallback mode faster than everyone else. They could do this to prevent someone posting a block containing the censored transactions.

These are essentially the same and rely on the assumption that you will be able to stop me from posting a transaction to the base-layer forced inclusion queue. It is not enough that you can make transaction quickly you need to make sure that I cannot get it on ethereum and I just need 1 tx within a period of days or even weeks for your plan to fail.

Also, it is a very expensive dos because you gotta dos the entirety of ethereum to keep it up. And if you were able to dos ethereum, then it have the same censorship guarantee as the baselayer 😎.



---

> What do you think about: 
> 1. A certain number of valid txs in the forced inclusion list that stay censored > $t$ time units causes fallback mode to activate, regardless of whether $T_{\textsf{fallback, enter}}$ passed or not. 
> 2. Once fallback mode has been activated due to censored transactions, the first few blocks submitted must include a certain % of the tx on the forced inclusion list (as of time just prior to fallback mode) to get accepted. 

Comment:
The forced inclusion will only support valid txs, we need to check before as no naysayers.

What do you gain from 1? As long as it could be included in a happy path I would very much prefer to not enter the pain of the fallback. The effect of this would be similar to reducing the time something can sit in the forced queue before it must be added. And since we have different ways we could end up in there, I would rather that we don't have many different rules, such that is essentially just becomes: committee is not doing their job, fallback to based.

2. The forced inclusion list is essentially a *forced ordering* of transactions that you **MUST** use, it is "forced". If it have 40 transactions to be included by slot x, you have to include either a subset of those 40 and nothing else or all 40 and some other transactions.

aminsammara (approved):
> Based fallback is for chain growth, forced inclusion is for censorship. 
Lasse, 2024

in-progress/8404-based-fallback.md
```diff
@@ -0,0 +1,191 @@
...
+After $T_{\textsf{fallback}, \textsf{enter}}$ time that would push us into the based fallback.
+At this point, anyone can push the blocks with the censored transactions.
+Even if the committee is trying to limit the throughput by only pushing the minimum blocks, $T_{\textsf{fallback}, \textsf{exit}}$ can then be picked such that a minimum throughput could be supported over the duration.
+The minimum throughput here under the assumption that either they will need to push minimum blocks as fast as they can, or someone else might push a bigger block exiting even more.
+
+### Changes to existing architecture
+
+The changes should mostly be on the L1 contract and the sequencer client, with a minor potential change for the rollup circuit (might simply not be necessary).
+
+- The L1 contract need to deal with the flows as above.
```
just-mitch, line 139: I suppose the rollup would need to keep track of the most recent "based proven block" as well as the most recent "committee proven block"?

in-progress/8404-based-fallback.md
```diff
@@ -0,0 +1,191 @@
...
+- The sequencer client need to know when fallback is activated such that it does not try to act as it normally would.
+  - I believe simply doing nothing is best here, as the fallbacks can be seen as a fully separate type of sequencer then.
+- Add a new "fallback" sequencer, which skips a lot of the checks of the current sequencer but mainly build and prove blocks.
+- Tee L1 contract/rollup circuit mix needs to support individual block proofs for this specific mode.
```
just-mitch, line 143: Could they just stop at the BlockRoot rollup? Or do an epoch of size 1?

in-progress/8404-based-fallback.md
LHerskind: If you see it actively as two separate modes, e.g., a flag, then you don't need to know the last based proven, as you can just use the flag + the last proven block as indicator.

in-progress/8404-based-fallback.md
LHerskind: Yes, the simplest would be just stopping at the block root and then publish that one 👍

just-mitch (commented):
Looks good, only question is what is the interface for triggering an exit of based mode? Does the committee just propose as normal and the rollup flips a switch if it has been long enough?

PR #28, opened 2024-09-17
Mt/forced-inclusion-updates

just-mitch:
Some minor cleanup.

My big question though is:
We don't actually need to store the historic forced pending tips and included pending for correctness right?

For example, we could just keep one `forced_pending_tip`, reset it to the `forced_inclusion_tip` on `prune`. Similarly keep one `included_pending` map, and clear it on `prune`. Is this really cheaper?

### DIFF
--- a/in-progress/8131-forced-inclusion.md
+++ b/in-progress/8131-forced-inclusion.md
@@ -304,7 +304,8 @@ Well, yes, but it might get confusing with both the pending and non pending func
 
 
 ```python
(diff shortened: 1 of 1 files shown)


LHerskind:
Adding the same comment here as discussed privately. 

If you are to clear a map it need to do as "resets" as there have been changes, e.g., scalers linearly. But by taking a new map, it is constant diff for the prune. This is a meaningful gas saving, and avoids a dos attack as well.

PR #29, opened 2024-10-17
Fees

just-mitch:


in-progress/8757-fees/design.md
```diff
@@ -0,0 +1,189 @@
...
+A transaction incurs the following costs:
+1. Proposers/validators simulate the transaction
+2. Proposers publish pending blocks to L1
+3. Provers generate the avm proofs for the transaction
+4. Provers generate the rollup proofs for the blocks/epoch
+5. Provers verify the epoch proof on L1
+
+Some costs are independent of the computation or data of the transaction itself, specifically numbers 2, 4, and 5; regardless, they must be covered by the transaction.
+
+## Protocol Defined Constants
```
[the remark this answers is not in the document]
LHerskind: Unrelated, but should we write these in all caps, for constants?

in-progress/8757-fees/design.md
```diff
@@ -0,0 +1,189 @@
...
+
+After simulating a block, the mana used by each transaction is known.
+
+A proposer can compute a block's `total_mana_used` by summing the `mana_used` fields of all transactions in the block.
+
+This `total_mana_used` will be part of the L2 block header.
+
+The rollup contract will maintain a `excess_mana` variable for the proven and pending chains.
+
+The `excess_mana` is computed from the previous block as the difference between `total_mana_used` and `target_mana_per_block`.
```
LHerskind, line 83: Not fully. It is not just the diff and the diff of the last it is an accumulated excess. If not accumulated we would max out much lower.

```python
spent = self.headers[-1].mana_used
excess = self.headers[-1].excess_mana
if excess + spent < self.mana_target:
    return 0
return excess + spent - self.mana_target
```

or I guess you can also write it as this
```python
diff = spent - self.mana_target
excess = max(0, excess + diff)
```

in-progress/8757-fees/design.md
```diff
@@ -0,0 +1,192 @@
...
+
+Compute the L2 congestion factor as the "fake exponential" of the ratio of `total_mana_used` plus the previous block's `excess_mana` to `target_mana_per_block`, and multiply by a constant factor of 1e9:
+
+```math
+\text{congestion factor} = \text{congestion factor multiplier} * \text{exp}\left(\frac{\text{total mana used} + \text{previous excess mana}}{\text{target mana per block}}\right)
```
LHerskind, line 88: - The thing you are calling the `congestion_factor_multiplier` is essentially the minimum congestion factor.
- It is not the target mana per block as the denominator, it is a `update_fraction` which can be computed from target mana but not directly. For example, if you want a max increment of 12%, you can use mana target divided by 0.117, e.g., if the total fraction is 0.117 that is an increment of 12% so for each entity total mana per block you are above it will add approx 12%.

in-progress/8757-fees/design.md
```diff
@@ -0,0 +1,192 @@
...
+```
+
+The base fee in wei/mana is then:
+
+```math
+\text{base wei per mana} = \text{congestion factor} * \text{cost of block in wei/mana}
+```
+
```
LHerskind, line 96: I would add the base fee in the fee asset here since you actually never mention it in here, and that is the thing that we will be feeding into the rollup.

in-progress/8757-fees/design.md
```diff
@@ -0,0 +1,192 @@
...
+The amount of mana a transaction consumes is:
+
+```math
+\text{mana}_{tx} = \text{overhead mana per tx} + \sum_{op \in tx} \text{mana per operation}(op)
+```
+
+Therefore, the cost of a transaction's proposal, data publication, and verification at time $i$ is, denominated in the fee asset:
+
+```math
+\text{fee asset}_{tx} = \text{mana}_{tx} * \text{base wei per mana} * \text{fee asset per wei}
```
LHerskind, line 109: With above changing, this is also just mana * mana base fee, simple

in-progress/8757-fees/design.md
```diff
@@ -0,0 +1,192 @@
...
+```
+
+## Distribution of the fee asset
+
+When a transaction is included in a block in the pending chain, the L2 balance of the fee asset for the `fee_payer` of the transaction is reduced by the transactions's fee asset cost.
+
+When an epoch is proven, the fees for the epoch are paid out to each proposer for the block(s) they proposed.
+
+A fixed percentage of each block's fee asset fees are paid to the prover of the epoch. The percentage is determined based on the quote submitted by the prover for the epoch.
```
LHerskind, line 118: I would not call it fixed, if it is based on the quote.

in-progress/8757-fees/design.md
```diff
@@ -0,0 +1,192 @@
...
+
+This quote is submitted by the prover in epoch `i+1` to prove epoch `i`, and is thus claimed/submitted to L1 by a proposer in epoch `i+1`.
+
+In order to prevent a prover bribing a proposer into accepting a quote with an extremely high fee, the maximum fee a prover can submit for an epoch proof quote is `maximum_epoch_proof_quote_fee`.
```
LHerskind, line 122: Related to the extremely high, one of my pr added a check to ensure you cannot require more than 100% hehe.

in-progress/8757-fees/design.md
```diff
@@ -0,0 +1,192 @@
...
+### `wei_per_l1_gas` and `wei_per_l1_blob_gas`
+
+The rollup contract will maintain an `wei_per_l1_gas` oracle. 
+
+Same for `wei_per_l1_blob_gas`.
+
+They will be updated up to every `underlying_base_fee_oracle_update_interval` slots.
+
+
+## Getting the cost of a block in wei/mana
```
LHerskind, line 62: I think we should be making a note around that this is the cost under the assumption that "x" mana is used. With that it is also a lot simpler to follow why some of these are constants even if not really constant.

in-progress/8757-fees/design.md
```diff
@@ -0,0 +1,192 @@
...
+
+## Clarifications and considerations
+
+### Why is `blobs_per_block` a constant?
+
+It makes computation easier. Otherwise block builder would need to know the number of blobs consumed by the transaction while it is building.
+
+### L1 Congestion
+
+If L1 gas spikes during an epoch, this will be reflected in the TST cost of a transaction through factoring in the higher `wei_per_l1_gas` and `wei_per_l1_blob_gas` into the TST cost.
```
LHerskind, line 132:
```suggestion
If L1 gas spikes during an epoch, this will be reflected in the `wei_per_l1_gas` and `wei_per_l1_blob_gas` oracles and make their way into the cost.
```

in-progress/8757-fees/design.md
```diff
@@ -0,0 +1,192 @@
...
+
+However, the base fee is computed at the time of a block being proposed, based on the current L1 base fees. These may change between the block being proposed and the epoch being verified.
```
LHerskind, line 134: It is not based on the current L1 base fees, but based on the oracle. So following, just not as quickly.

in-progress/8757-fees/design.md
```diff
@@ -0,0 +1,192 @@
...
+
+Thus, it is possible for L1 gas prices to spike to a point where it is not profitable for a prover to submit their proof to L1. There are two retorts:
+1. The prover should bake in a risk premium to compensate them for this risk; it is similar to the risk that the price of TST relative to any other asset shifts against them.
```
LHerskind, line 137: TST -> fee asset. Also, it is not really the same risk, as it is much more volatile than prices, unless we are speaking bottom of the barrel meme coins.

in-progress/8757-fees/design.md
```diff
@@ -0,0 +1,192 @@
...
+2. The prover can wait until the end of the epoch waiting for a less congested window to submit their proof to L1.
+
+### Not charging for DA separately
+
+Current thinking is that proving costs will dominate the cost of a transaction, so metering DA is not necessary.
```
LHerskind, line 142: We are metering it. We are just throwing it into the mix to just have one unit to account for now and not keeping two different costs ready.

in-progress/8757-fees/design.md
```diff
@@ -0,0 +1,265 @@
...
+
+A transaction incurs the following costs:
+1. Proposers/validators simulate the transaction
+2. Proposers publish pending blocks to L1
+3. Proposers/validators/nodes update their state
+4. Provers generate the avm proofs for the transaction
+5. Provers generate the rollup proofs for the blocks/epoch
+6. Provers verify the epoch proof on L1
+
+Some costs are independent of the computation or data of the transaction itself, specifically numbers 2, 4, and 5; regardless, they must be covered by the transaction.
```
LHerskind, line 25: The addition of the state updates (3) means that the below should also be updated to be 2, 5 and 6? 
```suggestion
Some costs are independent of the computation or data of the transaction itself, specifically numbers 2, 5, and 6; regardless, they must be covered by the transaction.
```

in-progress/8757-fees/design.md
```diff
@@ -0,0 +1,265 @@
...
+- `PROVING_COST_PRECISION` - the precision of the `proving_cost_modifier` (e.g. 1e9)
+- `MINIMUM_FEE_ASSET_PRICE` - the minimum price in wei of the fee asset (e.g. TST)
+- `MAXIMUM_FEE_ASSET_PRICE_PERCENT_CHANGE_PER_BLOCK` - the maximum percentage increase in the price of the fee asset per block (e.g. 1%)
+- `FEE_ASSET_PRICE_PRECISION` - the precision of the `fee_asset_price_modifier` (e.g. 1e9)
+- `UNDERLYING_BASE_FEE_ORACLE_UPDATE_INTERVAL` - the minimum number of slots between updates to the underlying base fee oracle (e.g. 4)
+- `MINIMUM_CONGESTION_MULTIPLIER` - the minimum value the congestion multiplier can take (e.g. 1)
+- `CONGESTION_MULTIPLIER_DAMPER` - the constant factor to dampen movement in the congestion multiplier (e.g. 8.547)
+- `MAXIMUM_EPOCH_PROOF_QUOTE_FEE` - the maximum basis point fee a prover can submit for an epoch proof quote (e.g. 9000)
+
+## Block Header Fields
```
LHerskind, line 46: Not clear to me from this if this is our real block header? Like, we don't need the excess ready for the circuits for example, and we kinda can get away with having for `proven`, `pending` and the last block in each `epoch`. The `base_fee_asset_per_mana` is something we already got as part of the `GasFee` of the global variables (don't recall exact name, think it is `l2_gas` or something like that).

in-progress/8757-fees/design.md
```diff
@@ -0,0 +1,265 @@
...
+
+## Block Execution Cost
+
+The cost of a block's execution in wei is equal to the following:
+```math
+\begin{aligned}
+\text{execution gas} &= \text{L1 GAS PER BLOCK PROPOSED}  \\
+&+ \text{BLOBS PER BLOCK} * \text{POINT EVALUATION PRECOMPILE GAS} \\
+&+ \frac{\text{L1 GAS PER EPOCH VERIFIED}}{\text{L2 SLOTS PER L2 EPOCH}} \\
+\text{execution wei} &= \text{execution gas} * \text{wei per l1 gas}
```
LHerskind, line 69: Down here, I would rather call it `execution_cost` to make it clearer that it is a cost. Kinda like, having a execution cost measured in usd, it is not execution_usd. Kinda a nit, but just clarity. Probably don't really need to specify the `execution_gas` can just go directly to the cost. Or, l1 costs at least.

in-progress/8757-fees/design.md
```diff
@@ -0,0 +1,265 @@
...
+## Rollup Contract Fields
+
+- `proving_cost_modifier` - a value used in the computation of the proving cost per mana
+- `proving_cost_modifier_delta` - an "oracle" used to update the `proving_cost_modifier` from the previous block
+- `fee_asset_price_modifier` - a value used in the computation of the fee asset price per wei
+- `fee_asset_price_modifier_delta` - an "oracle" used to update the `fee_asset_price_modifier` from the previous block
+- `wei_per_l1_gas` - the cost of L1 gas in wei. Updated by anyone to the current L1 gas price at most every `UNDERLYING_BASE_FEE_ORACLE_UPDATE_INTERVAL` slots
+- `wei_per_l1_blob_gas` - the cost of L1 blob gas in wei. Updated by anyone to the current L1 blob gas price at most every `UNDERLYING_BASE_FEE_ORACLE_UPDATE_INTERVAL` slots
+
+## Block Execution Cost
```
LHerskind, line 61: L1 execution cost?

in-progress/8757-fees/design.md
```diff
@@ -0,0 +1,265 @@
...
+
+They do so by updating the `proving_cost_modifier_delta` field, which is used to update the `proving_cost_modifier` stored in the rollup contract.
+
+That is,
+
+```math
+\text{new proving cost modifier} := \text{old proving cost modifier} + \text{proving cost modifier delta}
+```
+
+The new `proving_cost_modifier_delta` is capped at `MAXIMUM_PROVING_COST_PERCENT_CHANGE_PER_BLOCK` * `PROVING_COST_PRECISION`.
```
LHerskind, line 121: We cap it both + or -, it should also support reducing by 1% for example.

in-progress/8757-fees/design.md
```diff
@@ -0,0 +1,265 @@
...
+
+The takeaway is that, in this case, if we cap the change in the numerator between 0 and 1e9, the percent change is at most ~1%.
+
+### `proving_cost_wei_per_mana`
+
+```math
+\text{proving cost wei per mana} = \text{MINIMUM PROVING COST PER MANA} * \text{exp}\left(\frac{\text{proving cost modifier}}{100*\text{PROVING COST PRECISION}}\right)
+```
+
+A proposer can adjust the `proving_cost_modifier` stored in the rollup contract by up to `MAXIMUM_PROVING_COST_PERCENT_CHANGE_PER_BLOCK` each slot.
```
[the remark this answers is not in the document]
LHerskind: The validators can essentially use it to convey to the provers what they will be willing to pay. They could already increase the cost from the other "oracles", so there is an assumption there that they will essentially increase the prices as much as they can until users stop sending transactions. And in the case where they bump the proving a lot, I would expect to see the provers making bids for bigger shares essentially seeing it as a "we will pay you more". 

But tldr: I expect they will increase all the values they can influence as much as they can to make the fees as big as possible while still having people send transactions.

in-progress/8757-fees/design.md
```diff
@@ -0,0 +1,265 @@
...
+That is,
+
+```math
+\text{new proving cost modifier} := \text{old proving cost modifier} + \text{proving cost modifier delta}
+```
+
+The new `proving_cost_modifier_delta` is capped at `MAXIMUM_PROVING_COST_PERCENT_CHANGE_PER_BLOCK` * `PROVING_COST_PRECISION`.
+
+
+### `fee_asset_per_wei`
```
LHerskind, line 124: I think some of the naming makes it slightly confusing because it mixes things like units and precisions, that was part of the reason it was just `fee_asset_price` in the other.

just-mitch: I find "price" to be ambiguous though. I'd rather be explicit about the units and clear up the stuff with precision, which I think I did.

in-progress/8757-fees/design.md
just-mitch: I've made it very clear. I also kept the additional step to help with formatting.

in-progress/8757-fees/design.md
```diff
@@ -0,0 +1,265 @@
...
+- `L1_GAS_PER_BLOCK_PROPOSED` - the amount of L1 gas required to propose an L2 block on L1 (e.g. 0.2e6)
+- `L1_GAS_PER_EPOCH_VERIFIED` - the amount of L1 gas required to verify an L2 epoch on L1 (e.g. 1e6)
+- `MINIMUM_PROVING_COST_PER_MANA` - the minimum cost in wei for proving a unit of mana (e.g. 100)
+- `MAXIMUM_PROVING_COST_PERCENT_CHANGE_PER_BLOCK` - the maximum percentage increase in the cost of proving a unit of mana per block (e.g. 1%)
+- `PROVING_COST_PRECISION` - the precision of the `proving_cost_modifier` (e.g. 1e9)
+- `MINIMUM_FEE_ASSET_PRICE` - the minimum price in wei of the fee asset (e.g. TST)
+- `MAXIMUM_FEE_ASSET_PRICE_PERCENT_CHANGE_PER_BLOCK` - the maximum percentage increase in the price of the fee asset per block (e.g. 1%)
+- `FEE_ASSET_PRICE_PRECISION` - the precision of the `fee_asset_price_modifier` (e.g. 1e9)
+- `UNDERLYING_BASE_FEE_ORACLE_UPDATE_INTERVAL` - the minimum number of slots between updates to the underlying base fee oracle (e.g. 4)
+- `MINIMUM_CONGESTION_MULTIPLIER` - the minimum value the congestion multiplier can take (e.g. 1)
```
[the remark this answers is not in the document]
just-mitch: I think this one is okay now. Based on how this is used, `1` makes sense here, and I cleared up the precision stuff to use naming in line with EIP-4844.

in-progress/8757-fees/design.md
```diff
@@ -0,0 +1,189 @@
...
+4. Provers generate the rollup proofs for the blocks/epoch
+5. Provers verify the epoch proof on L1
+
+Some costs are independent of the computation or data of the transaction itself, specifically numbers 2, 4, and 5; regardless, they must be covered by the transaction.
+
+## Protocol Defined Constants
+
+- `overhead_mana_per_tx` - the overhead cost in mana for a transaction
+- `target_mana_per_block` - the amount of mana that an "average" block is expected to consume
+- `blobs_per_block` - the number of blobs per block that are compensated for in the base fee
```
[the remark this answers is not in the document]
just-mitch: Actually this will get especially rolled into the base fee because we straight up calculate the current^ cost of the blobs in wei and roll that into the base fee.

^current as of the last `UNDERLYING_BASE_FEE_ORACLE_UPDATE_INTERVAL`

in-progress/8757-fees/design.md
```diff
@@ -0,0 +1,270 @@
...
+\text{new proving cost modifier} := \text{old proving cost modifier} + \text{proving cost modifier delta}
+```
+
+The new `proving_cost_modifier_delta` is capped at (+/-) `MAXIMUM_PROVING_COST_PERCENT_CHANGE_PER_BLOCK` * `PROVING_COST_UPDATE_FRACTION` / 100.
+
+
+### `fee_asset_per_wei`
+
+```math
+\text{fee asset per wei} = \text{MINIMUM FEE ASSET PRICE} * \text{exp}\left(\frac{\text{fee asset price modifier}}{\text{FEE ASSET PRICE UPDATE FRACTION}}\right)
```
aminsammara, line 130: How do proposers decrease fee asset per wei? 

i.e. once there is an update that takes `fee asset per wei` above the minimum constant value, what update could the proposers make to decrease it?

aminsammara: Ok nvm, this isn't in the form new value = old value * exp... woops

in-progress/8757-fees/design.md
```diff
@@ -0,0 +1,270 @@
...
+
+- `OVERHEAD_MANA_PER_TX` - the overhead cost in mana for a transaction (e.g. 21_000)
+- `TARGET_MANA_PER_BLOCK` - the amount of mana that an "average" block is expected to consume (e.g. 15e6)
+- `MAXIMUM_MANA_PER_BLOCK` - the maximum amount of mana that a block can consume (e.g. 2 * TARGET_MANA_PER_BLOCK)
+- `BLOBS_PER_BLOCK` - the number of blobs per block that are compensated for in the base fee (e.g. 3)
+- `L2_SLOTS_PER_L2_EPOCH` - the number of L2 slots in each L2 epoch (e.g. 32)
+- `L1_GAS_PER_BLOCK_PROPOSED` - the amount of L1 gas required to propose an L2 block on L1 (e.g. 0.2e6)
+- `L1_GAS_PER_EPOCH_VERIFIED` - the amount of L1 gas required to verify an L2 epoch on L1 (e.g. 1e6)
+- `MAXIMUM_EPOCH_PROOF_QUOTE_FEE` - the maximum basis point fee a prover can submit for an epoch proof quote (e.g. 9000)
+- `UNDERLYING_BASE_FEE_ORACLE_UPDATE_INTERVAL` - the minimum number of slots between updates to the underlying base fee oracle (e.g. 4)
```
spalladino, line 37: Why is there a minimum update interval? Wouldn't it be easier if the act of proposing a block automatically updates these oracles for the next block?

in-progress/8757-fees/design.md
```diff
@@ -0,0 +1,270 @@
...
+
+When a transaction is included in a block in the pending chain, the L2 balance of the fee asset for the `fee_payer` of the transaction is reduced by the transactions's fee asset cost.
+
+When an epoch is proven, the fees for the epoch are paid out to each proposer for the block(s) they proposed.
+
+A percentage of each block's fee asset fees are paid to the prover of the epoch. The percentage is determined based on the quote submitted by the prover for the epoch.
+
+This quote is submitted by the prover in epoch `i+1` to prove epoch `i`, and is thus claimed/submitted to L1 by a proposer in epoch `i+1`.
+
+In order to prevent a prover bribing a proposer into accepting a quote with an extremely high fee, the maximum fee a prover can submit for an epoch proof quote is `MAXIMUM_EPOCH_PROOF_QUOTE_FEE`.
```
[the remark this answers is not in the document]
aminsammara: We're thinking about this and the best current solution is for `MAXIMUM_EPOCH_PROOF_QUOTE_FEE` to track the fair price of proving, such that sequencers should not "care" which quote they pick. They should just care that they picked a quote. 

Sequencers don't get rewarded for picking a quote.

in-progress/8757-fees/design.md
```diff
@@ -0,0 +1,270 @@
...
+\text{fee asset}_{tx} = \text{mana}_{tx} * \text{base fee asset per mana}
+```
+
+## Distribution of the fees
+
+When a transaction is included in a block in the pending chain, the L2 balance of the fee asset for the `fee_payer` of the transaction is reduced by the transactions's fee asset cost.
+
+When an epoch is proven, the fees for the epoch are paid out to each proposer for the block(s) they proposed.
+
+A percentage of each block's fee asset fees are paid to the prover of the epoch. The percentage is determined based on the quote submitted by the prover for the epoch.
```
aminsammara, line 196: Just checking this includes block rewards right? 

Otherwise the network could not subsidize proving, users would have to pay the full proving cost. 

We could have: 
1) MAXIMUM_EPOCH_PROOF_QUOTE_FEE = % (block fees + block rewards) or
2) provers always get `proving wei per mana  * total_mana_per_block`  + MAXIMUM_EPOCH_PROOF_QUOTE_FEE% * total_block_reward.

in-progress/8757-fees/design.md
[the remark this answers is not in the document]
spalladino: If we did that, maybe we can change the semantics of the quote that provers send, so it's a `minimumBasisPointsFee` rather than a `basisPointsFee`, and the protocol dictates what they are actually paid based on this fair price?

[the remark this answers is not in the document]
just-mitch: My thought here was that proposers in epoch `N` could signal the max fee they're willing to accept, then we take the average in the rollup contract, and the claimed bid must be less than it.

in-progress/8757-fees/design.md
```diff
@@ -0,0 +1,270 @@
...
+|                      |                                                                    |
+| -------------------- | ------------------------------------------------------------------ |
+| Issue                | https://github.com/AztecProtocol/aztec-packages/issues/8757        |
+| Owners               | @just-mitch @LHerskind                                             |
+| Approvers            | @Maddiaa0 @PhilWindle @dbanks12 @nventuro @joeandrews @aminsammara |
+| Target Approval Date | 2024-10-24                                                         |
+
+## Mana and Costs
+
+Refer to the unit of work on L2 as "mana", analogous to gas on L1. This has been referred to as "L2 gas" in the past.
```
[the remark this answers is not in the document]
just-mitch: Just to not have to always disambiguate L2 gas versus L1 gas. There's also an argument that "it's not an evm, so it's not gas". I'm definitely not beholden to the name.

in-progress/8757-fees/design.md
just-mitch: We could make the interval the same as the slot length, but we can't *only* do it as part of block proposal, since we want it  to be possible to be updated even when there are many missed slots.

in-progress/8757-fees/design.md
just-mitch: Block rewards were intentionally not discussed because they are not part of the fee mechanism.

What is implemented today on master is that the prover gets paid in Fee Asset:
```math
\text{proofClaim.basisPointFee} \times \sum_{\text{block} \in \text{epoch}} \text{Block Reward} + \text{block.fees} 
```

in-progress/8757-fees/design.md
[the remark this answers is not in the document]
rahul-kothari: I think everyone externally will use the phrase "L2 gas" when thinking about it

> it's not an evm, so it's not gas"
Should probably not use `wei` then either?

in-progress/8757-fees/design.md
```diff
@@ -0,0 +1,270 @@
...
+Thus, it is possible for L1 gas prices to spike to a point where it is not profitable for a prover to submit their proof to L1. There are two retorts:
+1. The prover should bake in a risk premium to compensate them for this risk; it is similar to the risk that the price of fee asset relative to any other asset shifts against them.
+2. The prover can wait until the end of the epoch waiting for a less congested window to submit their proof to L1.
+
+### Not charging for DA separately
+
+By effectively charging each proposed block as though it uses BLOBS_PER_BLOCK blobs, we can simplify by not charging separately for DA, and instead add the use of the blobs directly to the block cost.
+
+Regardless, the AVM supports DA gas metering per opcode, so it would not be difficult to add in the future if we change our mind.
+
```
dbanks12, line 223: Is this the only way this might affect the AVM? Or are the other ways as well?

in-progress/8757-fees/design.md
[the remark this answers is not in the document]
just-mitch: `wei` as referenced in this doc is real `wei` as referenced in L1.

Regardless, we need to have a separate meeting to decide this naming stuff once and for all.

in-progress/8757-fees/design.md
just-mitch: Yes, should be. I don't think any changes are needed with respect to the current Mana/L2Gas

in-progress/8757-fees/design.md
```diff
@@ -0,0 +1,189 @@
...
+| Issue                | https://github.com/AztecProtocol/aztec-packages/issues/8757        |
+| Owners               | @just-mitch @LHerskind                                             |
+| Approvers            | @Maddiaa0 @PhilWindle @dbanks12 @nventuro @joeandrews @aminsammara |
+| Target Approval Date | 2024-10-24                                                         |
+
+## Mana and Costs
+
+Refer to the unit of work on L2 as "mana", analogous to gas on L1.
+
+A transaction incurs the following costs:
```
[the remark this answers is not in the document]
iAmMichaelConnor: [Seeing this as resolved] What was the resolution to this point?

in-progress/8757-fees/design.md
```diff
@@ -0,0 +1,270 @@
...
+
+This quote is submitted by the prover in epoch `i+1` to prove epoch `i`, and is thus claimed/submitted to L1 by a proposer in epoch `i+1`.
+
+In order to prevent a prover bribing a proposer into accepting a quote with an extremely high fee, the maximum fee a prover can submit for an epoch proof quote is `MAXIMUM_EPOCH_PROOF_QUOTE_FEE`.
+
+## Clarifications and considerations
+
+### Why is `blobs_per_block` a constant?
+
+It makes computation easier. Otherwise block builder would need to know the number of blobs consumed by the transaction while it is building.
```
iAmMichaelConnor, line 206: Oh, this should've appeared earlier

iAmMichaelConnor: Wait, but blobs have a significant impact on the rollup's costs (or they will do once more people use blobs), so I'm not sure this makes sense to me.

iAmMichaelConnor: Please could you elaborate further on why it needs to be constant, to help me understand?
If I'm a block builder, I see a tx, I simulate it, I see how much data it wants to emit, I compute the L1 gas costs of that data (based on the proportion of blob(s) that this tx would use-up). I now know a rough estimate of how much ETH the data in this tx will cost the prover. Can we not then compute some approximation of this eth that the user can pay through L2 tokens?

in-progress/8757-fees/design.md
```diff
@@ -0,0 +1,270 @@
...
+
+However, these oracles lag the current L1 base fees. Furthermore, the base fee computed at the time of a block being proposed may differ from the base fee at the time of the epoch being verified.
+
+Thus, it is possible for L1 gas prices to spike to a point where it is not profitable for a prover to submit their proof to L1. There are two retorts:
+1. The prover should bake in a risk premium to compensate them for this risk; it is similar to the risk that the price of fee asset relative to any other asset shifts against them.
+2. The prover can wait until the end of the epoch waiting for a less congested window to submit their proof to L1.
+
+### Not charging for DA separately
+
+By effectively charging each proposed block as though it uses BLOBS_PER_BLOCK blobs, we can simplify by not charging separately for DA, and instead add the use of the blobs directly to the block cost.
```
iAmMichaelConnor, line 220: Also should've been nearer the top :)

in-progress/8757-fees/design.md
```diff
@@ -0,0 +1,270 @@
...
+
+Regardless, the AVM supports DA gas metering per opcode, so it would not be difficult to add in the future if we change our mind.
+
+Further, it is assumed that the proving cost will dominate the cost of the transaction, so this simplification makes implementation easier, and UX better.
```
iAmMichaelConnor, line 224: I might argue that in 5-10 years, the DA costs will dominate. There are provers that are already 10x faster than our public provers. So it'll be more like 100x+ faster in 5-10 years. Then DA will dominate.

in-progress/8757-fees/design.md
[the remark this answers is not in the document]
iAmMichaelConnor: This sounds like a fundamental problem with our incentive mechanisms? 
Or, at least, shouldn't this doc have more revisions before it can be merged?

iAmMichaelConnor (commented):
Sorry, lots of pedantic comments, from the pov of someone who _should_ be able to understand all of this, but is struggling to catch up. I went over-the-top on naming pedantry; sorry.
I think this doc might be missing (near the start) some background on your _journey_ to reach this proposal. It's just stating facts, and my brain keeps asking "why", "where did this come from?", "how did we get here?". I'm not saying the doc is wrong - you're all clever chaps - I'm saying the doc hasn't enabled me to fully catch up.
It's also missing an initial high-level vision / summary of what each actor will need to do and comprehend:
- A user would preferably only need to comprehend ... .
- A wallet builder (who abstracts fees for the user) will need to comprehend / compute...
- A sequencer/builder will need to comprehend / compute... . They will compute fees like this... . They will be paid fees in ... at time ... by ... .
- A prover will need to comprehend / compute ... . They will compute fees like this... . They will be paid fees in ... at time ... by ... .

in-progress/8757-fees/design.md
iAmMichaelConnor: Even in the shorter-term, with the number of rollups coming online, blob costs will go up.

in-progress/8757-fees/design.md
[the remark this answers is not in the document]
joeandrews: I missed this thread. I do think sequencers need to set in their proposal the minimum % fees they will accept for their block in epoch N. The proof quote in epoch N+1 must meet this constraint.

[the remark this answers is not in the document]
joeandrews: I think its fine in general for the sequencer in epoch N +1 to not be incentivised with a fee. Their incentive is to pick a proof quote that is most likely to succeed so their block in epoch N + 1 does not get re-orged.

in-progress/8757-fees/design.md
```diff
@@ -0,0 +1,270 @@
...
+6. Provers verify the epoch proof on L1
+
+Some costs are independent of the computation or data of the transaction itself, specifically numbers 2, 5, and 6; regardless, they must be covered by the transaction.
+
+## Protocol Defined Constants
+
+- `OVERHEAD_MANA_PER_TX` - the overhead cost in mana for a transaction (e.g. 21_000)
+- `TARGET_MANA_PER_BLOCK` - the amount of mana that an "average" block is expected to consume (e.g. 15e6)
+- `MAXIMUM_MANA_PER_BLOCK` - the maximum amount of mana that a block can consume (e.g. 2 * TARGET_MANA_PER_BLOCK)
+- `BLOBS_PER_BLOCK` - the number of blobs per block that are compensated for in the base fee (e.g. 3)
```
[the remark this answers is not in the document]
just-mitch: As noted elsewhere, this is just the number of blobs that are compensated by the base fee.

in-progress/8757-fees/design.md
```diff
@@ -0,0 +1,270 @@
...
+- `L2_SLOTS_PER_L2_EPOCH` - the number of L2 slots in each L2 epoch (e.g. 32)
+- `L1_GAS_PER_BLOCK_PROPOSED` - the amount of L1 gas required to propose an L2 block on L1 (e.g. 0.2e6)
+- `L1_GAS_PER_EPOCH_VERIFIED` - the amount of L1 gas required to verify an L2 epoch on L1 (e.g. 1e6)
+- `MAXIMUM_EPOCH_PROOF_QUOTE_FEE` - the maximum basis point fee a prover can submit for an epoch proof quote (e.g. 9000)
+- `UNDERLYING_BASE_FEE_ORACLE_UPDATE_INTERVAL` - the minimum number of slots between updates to the underlying base fee oracle (e.g. 4)
+- `MINIMUM_PROVING_COST_PER_MANA` - the minimum cost in wei for proving a unit of mana (e.g. 100)
+- `MAXIMUM_PROVING_COST_PERCENT_CHANGE_PER_BLOCK` - the maximum percentage increase in the cost of proving a unit of mana per block (e.g. 1%)
+- `PROVING_COST_UPDATE_FRACTION` - a value used to update the `proving_cost_modifier` (e.g. 1e11)
+- `MINIMUM_FEE_ASSET_PRICE` - the minimum price in wei of the fee asset (e.g. TST)
```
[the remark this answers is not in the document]
just-mitch: We just need a non-zero value that the market can scale off. It can be very small.

in-progress/8757-fees/design.md
```diff
@@ -0,0 +1,270 @@
...
+- `total_mana_used` - the total mana used by the block
+- `base_fee_asset_per_mana` - the base fee in fee asset per mana
+
+## Rollup Contract Fields
+
+The rollup contract contains the following fields:
+
+- `proving_cost_modifier` - a value used in the computation of the proving cost per mana
+- `proving_cost_modifier_delta` - an "oracle" used to update the `proving_cost_modifier` from the previous block
+- `fee_asset_price_modifier` - a value used in the computation of the fee asset price per wei
```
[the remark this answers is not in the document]
just-mitch: We tally our costs in wei, and then convert to TST with the aid of this multiplier.

in-progress/8757-fees/design.md
```diff
@@ -0,0 +1,270 @@
...
+
+The cost of a block's execution in wei is equal to the following:
+```math
+\begin{aligned}
+\text{execution cost in gas} &= \text{L1 GAS PER BLOCK PROPOSED}  \\
+&+ \text{BLOBS PER BLOCK} * \text{POINT EVALUATION PRECOMPILE GAS} \\
+&+ \frac{\text{L1 GAS PER EPOCH VERIFIED}}{\text{L2 SLOTS PER L2 EPOCH}} \\
+\text{execution cost in wei} &= \text{execution cost in gas} * \text{wei per l1 gas}
+\end{aligned}
+```
```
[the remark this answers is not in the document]
just-mitch: Blobs per block is fixed at a constant because to hit 10 TPS we likely need 3 blobs per L2 block.

Missed slots shouldn't change the math above. We divide by that constant to have each block pay "its share" of of verification.

in-progress/8757-fees/design.md
```diff
@@ -0,0 +1,270 @@
...
+
+This final value is the `base_fee_asset_per_mana` field in the L2 block header.
+
+## The cost of a transaction in the fee asset
+
+The amount of mana a transaction consumes is:
+
+```math
+\text{mana}_{tx} = \text{OVERHEAD MANA PER TX} + \sum_{op \in tx} \text{mana per operation}(op)
```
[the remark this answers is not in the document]
just-mitch: mana per operation is effectively the exact L2 gas metering we have in place today.

So an "operation" here is emitting data like note hashes and nullifiers in private (see noir-projects/noir-protocol-circuits/crates/private-kernel-lib/src/components/tail_to_public_output_composer/meter_gas_used.nr), as well as AVM metering.

The mana consumed by those operations ought to compensate for the various costs specific to that transaction (simulating, gossiping, proving AVM circuits, etc.)

OVERHEAD_MANA is to cover literally everything else, but mainly that transactions "share" in proving and verifying the rollup circuits.

in-progress/8757-fees/design.md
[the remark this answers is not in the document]
just-mitch: How about then we axe this sentence altogether and add this "average" functionality in a followup?

PR #32, opened 2024-11-02
Update 8404-based-fallback.md

mdqst:
Снимок экрана 2024-11-02 в 14 05 18

A typo was found and corrected.
Correct spelling is "**categories**".


just-mitch:
Thanks, but we don't accept spelling corrections from first time contributors.

PR #47, opened 2025-01-10
Proposer Forwarder Contract

just-mitch:
Add a forwarder contract that allows the sequencer client to take multiple actions in the same L1 transaction.

Adjust the sequencer client to batch its actions into a single L1 transaction.

### DIFF
--- a/in-progress/11103-forwarder-contract/design.md
+++ b/in-progress/11103-forwarder-contract/design.md
@@ -4,8 +4,8 @@
 | -------------------- | ---------------------------------------------------------------------------------- |
 | Issue                | [Forwarder Contract](https://github.com/AztecProtocol/aztec-packages/issues/11103) |
 | Owners               | @just-mitch                                                                        |
-| Approvers            | @LHerskind @PhilWindle                                                             |
-| Target Approval Date | 2025-01-20                                                                         |
+| Approvers            | @LHerskind @PhilWindle @spalladino @spypsy                                         |
+| Target Approval Date | 2025-01-15                                                                         |
 
 ## Executive Summary
 
@@ -15,9 +15,9 @@ Adjust the sequencer client to batch its actions into a single L1 transaction.
 
 ## Introduction
 
-Within the same L1 transaction, we cannot make blob transactions and regular transactions from the same address.
+Within the same L1 transaction, one cannot make blob transactions and regular transactions from the same address.
 
-We must be able to do that though, since we want to be able to do things like:
+However, aztec node operators must be able to do things like:
 
 - propose and l2 block
 - vote in the governance proposer contract
@@ -66,6 +66,10 @@ contract Forwarder is Ownable {
 }
 ```
 
+Note: this requires all the actions to succeed, so the sender must be sure that, e.g. a failed governance vote will not prevent the L2 block from being proposed.
+
+Note: this implementation is not technically part of the protocol, and as such will live in `l1-contracts/src/periphery`.
+
 ### Refactoring L1 Publisher
 
 L1 publisher will be broken into two classes:
@@ -100,11 +104,48 @@ The `Sequencer` uses its `L1TxManager` to make calls to:
 - cast a slashing vote
 - claim an epoch proof quote
 
+The return type of each of these functions will be a `Promise<TransactionReceipt>`, which will be resolved when the bundled transaction is included in an L1 block.
+
 These requests will be added to the `queuedRequests` list.
 
-The work loop will wait for a configurable amount of time (e.g. 6 seconds) into each L1 slot.
+The work loop will wait for a configurable amount of time (e.g. 6 seconds) into each L1 slot. This will be exposed as an environment variable `sequencer.l1TxSubmissionDeadline`.
 
 If there are any queued requests, it will send them to the forwarder contract, and flush the `queuedRequests` list.
+
+### L1PublishBlockStats
+
+Since one `L1PublishBlockStats` can be used for multiple actions, its constituent `event` will be changed to `actions`, which will be a `string[]`.
+
+The sequencer's `L1TxManager` and the prover node's `L1TxPublisher` will populate this array and record the metric.
+
+### Cancellation/Resend
+
+A complication is that ethereum nodes make replacement of blob transactions expensive, and cancelation impossible, as they operate under the assumption that rollups seldom/never need to replace/cancel blob transactions.
+
+See [geth's blob pool](https://github.com/ethereum/go-ethereum/blob/581e2140f22566655aa8fb2d1e9a6c4a740d3be1/core/txpool/blobpool/blobpool.go) for details/constraints.
+
+This is not true for Aztec's decentralized sequencer set with strict L1 timeliness requirements on L2 blocks.
+
+So a concern is the following scenario:
+
+- proposer A submits a tx with nonce 1 (with a blob) that is not priced aggressively enough
+- Tx1 sits in the blob pool, but is not included in an L1 block
+- proposer A tries to submit another transaction, but needs to know to use Tx2
+- Tx1 needs to be replaced with a higher fee, but it will revert if the network is in a different L2 slot and the bundle contained a proposal
+
+This is addressed by:
+
+- Upgrading viem to at least v2.15.0 to use their nonceManager to be aware of pending nonces
+- Aggressive pricing of blob transactions
+- The L1TxUtils will be able to speed up Tx1 (even if it reverts), which should unblock Tx2
+
+A different approach would be to have the sequencer client maintain a pool of available forwarder contracts, and use one until it gets stuck, then switch to the next one: presumably by the time the sequencer client gets to the original forwarder contract, the blob pool will have been cleared.
+
+Broader changes about changing the timeliness requirements are not in scope for this change.
+
+### Setup
+
+There will be an optional environment variable `sequencer.forwarderContractAddress` that will be used to specify the forwarder contract address. To improve UX, there will be a separate environment variable `sequencer.deployForwarderContract` that will default to `true` and be used to specify whether the forwarder contract should be deployed. If so, the Aztec Labs implementation of the forwarder contract will be deployed and used by the sequencer.
 
 ### Gas
 
@@ -139,35 +180,23 @@ Fill in bullets for each area that will be affected by this change.
 - [ ] Aztec.nr
 - [ ] Enshrined L2 Contracts
 - [ ] Private Kernel Circuits
-- [ ] Sequencer
+- [x] Sequencer
 - [ ] AVM
 - [ ] Public Kernel Circuits
 - [ ] Rollup Circuits
-- [ ] L1 Contracts
-- [ ] Prover
+- [x] L1 Contracts
+- [x] Prover
 - [ ] Economics
 - [ ] P2P Network
 - [ ] DevOps
 
 ## Test Plan
 
-Outline what unit and e2e tests will be written. Describe the logic they cover and any mock objects used.
+The primary test is [cluster governance upgrade](https://github.com/AztecProtocol/aztec-packages/issues/9638), ensuring that block production does not stall (as it currently does).
 
 ## Documentation Plan
 
-Identify changes or additions to the user documentation or protocol spec.
-
-## Rejection Reason
-
-If the design is rejected, include a brief explanation of why.
-
-## Abandonment Reason
-
-If the design is abandoned mid-implementation, include a brief explanation of why.
-
-## Implementation Deviations
-
-If the design is implemented, include a brief explanation of deviations to the original design.
+No plans to document this as yet: the node operator guide effectively does not exist.
 
 ## Disclaimer
 
--- a/in-progress/11103-forwarder-contract/design.md
+++ b/in-progress/11103-forwarder-contract/design.md
@@ -70,27 +70,30 @@ contract Forwarder is Ownable {
 
 L1 publisher will be broken into two classes:
 
-- `SequencerL1API`
-- `ProverL1API`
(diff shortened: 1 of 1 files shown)


in-progress/11103-forwarder-contract/design.md
```diff
@@ -0,0 +1,203 @@
...
+- knowledge of the sequencer's forwarder contract
+- knowledge of L1 slot boundaries
+- knowledge of successful L1 transactions per L2 slot
+
+The `Sequencer` uses its `L1TxManager` to make calls to:
+
+- propose an l2 block
+- cast a governance proposal vote
+- cast a slashing vote
+- claim an epoch proof quote
```
spalladino, line 105: Have we considered **not** bundling votes with block proposals? If we kept at least two different EOAs, we could push the two txs in parallel, with the voting one being "lighter" and easier to bump/replace (since it's blobless).

I guess this depends on whether votes need to happen strictly during the proposer slot.

in-progress/11103-forwarder-contract/design.md
```diff
@@ -0,0 +1,203 @@
...
+Note: this implementation is not technically part of the protocol, and as such will live in `l1-contracts/src/periphery`.
+
+### Refactoring L1 Publisher
+
+L1 publisher will be broken into two classes:
+
+- within `@aztec/sequencer-client`, there will be a `L1TxManager`
+- within `@aztec/prover-node`, there will be a `L1TxPublisher`
+
+Under the hood, both of these will use the `L1TxUtils` to create and send L1 transactions.
```
spalladino, line 80: Note that the `L1Publisher` is being used also for querying (we were lazy, yeah), which should not be among its responsibilities. I added a wrapper around the Rollup contract for querying some info at `yarn-project/ethereum/src/contracts/rollup.ts`, maybe we can move all getters there?

spalladino (commented):
Thanks for putting this together @just-mitch! Left a few comments.

in-progress/11103-forwarder-contract/design.md
```diff
@@ -0,0 +1,203 @@
...
+- Aggressive pricing of blob transactions
+- The L1TxUtils will be able to speed up Tx1 (even if it reverts), which should unblock Tx2
+
+A different approach would be to have the sequencer client maintain a pool of available forwarder contracts, and use one until it gets stuck, then switch to the next one: presumably by the time the sequencer client gets to the original forwarder contract, the blob pool will have been cleared.
+
+Broader changes about changing the timeliness requirements are not in scope for this change.
+
+### Setup
+
+There will be an optional environment variable `sequencer.forwarderContractAddress` that will be used to specify the forwarder contract address. To improve UX, there will be a separate environment variable `sequencer.deployForwarderContract` that will default to `true` and be used to specify whether the forwarder contract should be deployed. If so, the Aztec Labs implementation of the forwarder contract will be deployed and used by the sequencer.
```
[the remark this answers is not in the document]
just-mitch: Agreed, but I think the user should be able to specify their own forwarder implementation. Maybe we should rename this optional environment variable to `sequencer.customForwarderContractAddress`, and if it is not set then we automatically deploy one for them (using UDD based on their SEQ_PUBLISHER_PRIVATE_KEY).

in-progress/11103-forwarder-contract/design.md
```diff
@@ -0,0 +1,203 @@
...
+- proposer A tries to submit another transaction, but needs to know to use Tx2
+- Tx1 needs to be replaced with a higher fee, but it will revert if the network is in a different L2 slot and the bundle contained a proposal
+
+This is addressed by:
+
+- Upgrading viem to at least v2.15.0 to use their nonceManager to be aware of pending nonces
+- Aggressive pricing of blob transactions
+- The L1TxUtils will be able to speed up Tx1 (even if it reverts), which should unblock Tx2
+
+A different approach would be to have the sequencer client maintain a pool of available forwarder contracts, and use one until it gets stuck, then switch to the next one: presumably by the time the sequencer client gets to the original forwarder contract, the blob pool will have been cleared.
```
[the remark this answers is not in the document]
just-mitch: Yes, it was a typo. But I still think we might want the forwarder contract in this case; the forwarder would be registered as the proposer of the validator, and the different EOAs would just be "owners" of the forwarder contract. Is that what you're imagining?

in-progress/11103-forwarder-contract/design.md
just-mitch: Yes, the vote needs to come from the proposer address within its slot.

We could have a validator specify two addresses, one that will submit blocks, and another that will do everything else. 

But if we do the pooled EOAs to abandon blob transactions, I still think we'd want the forwarder contract as a layer of indirection, otherwise on L1 we'd need to walk through all the approved EOAs.

I think its simpler to keep the single proposer address that does everything.

spalladino: > But if we do the pooled EOAs to abandon blob transactions, I still think we'd want the forwarder contract as a layer of indirection, otherwise on L1 we'd need to walk through all the approved EOAs.

I was thinking about just using the forwarder contract address as "the" validator's address in the staking registry. But I didn't consider that we're requiring off-chain signatures for attestations, so we cannot use the forwarder contract as it doesn't have keys associated to it (damn lack of AA).

in-progress/11103-forwarder-contract/design.md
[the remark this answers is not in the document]
spalladino: > But I still think we might want the forwarder contract in this case; the forwarder would be registered as the proposer of the validator, and the different EOAs would just be "owners" of the forwarder contract. Is that what you're imagining?

Exactly, with the caveat that the forwarder cannot sign (which I only now realised) because it doesn't have keys associated to it.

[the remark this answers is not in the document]
just-mitch: I'll include/clarify this.

in-progress/11103-forwarder-contract/design.md
```diff
@@ -0,0 +1,201 @@
...
+The `L1TxManager` will have:
+
+- `queuedRequests: L1TxRequest[]`
+- knowledge of the sequencer's forwarder contract
+
+It will have an interface of:
+
+```typescript
+interface L1TxManager {
+  addRequest(request: L1TxRequest): void;
```
LHerskind, line 103: What about a "clear"? Based on time of execution, it might not actually make sense for you to send it so better to drop then send a tx you know will fail.

in-progress/11103-forwarder-contract/design.md
```diff
@@ -0,0 +1,201 @@
...
+
+This is addressed by:
+
+- Upgrading viem to at least v2.15.0 to use their nonceManager to be aware of pending nonces
+- Aggressive pricing of blob transactions
+- The L1TxUtils will be able to speed up Tx1 (even if it reverts), which should unblock Tx2
+
+### Setup
+
+There will be an optional environment variable `sequencer.custForwarderContractAddress` that can be used to specify a custom forwarder contract address.
```
LHerskind, line 140: Is there a reason that `custom` is shorthanded to `cust`?

in-progress/11103-forwarder-contract/design.md
```diff
@@ -0,0 +1,201 @@
...
+
+If this is not set, the sequencer will deploy the Aztec Labs implementation of the forwarder contract, using the Universal Deterministic Deployer, supplying the sequencer's address as the deployment salt, and the sequencer's address as the owner.
+
+### Gas
+
+Once the L2 block body is removed from calldata, the "static" arguments to call the propose function should be under 1KB.
```
LHerskind, line 146: I know you say static here, but signatures still are being passed around.
`65 * (128 * 2 / 3) + 128 / 3 * 32` ~7kb. 

Also,

in-progress/11103-forwarder-contract/design.md
```diff
@@ -0,0 +1,201 @@
...
+
+Operating at 10TPS, this would mean an overhead of under 3 gas per L2 transaction.
+
+Unfortunately, the current design to support forced inclusion requires a hash for each transaction in the proposal args.
```
LHerskind, line 150: My alzheimers is kicking in, but is this actually true? As I recall it, the "hash of each transaction" is ONE hash of all of them, e.g., `txs_hash` being a commitment to all the transactions in the block, e.g., 32 bytes for the full block.

in-progress/11103-forwarder-contract/design.md
```diff
@@ -0,0 +1,201 @@
...
+- propose and l2 block
+- vote in the governance proposer contract
+- claim an epoch proof quote
+  all in the same L1 block.
+
+### Goals
+
+- Allow the sequencer client to take multiple actions in the same L1 transaction
+- No changes to governance/staking
+- Under 50 gas overhead per L2 transaction when operating at 10TPS
```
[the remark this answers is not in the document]
just-mitch: It was finger in the air, but I think this layer of indirection shouldn't eat more than 10% of our 500 gas target.

in-progress/11103-forwarder-contract/design.md
```diff
@@ -0,0 +1,201 @@
...
+
+## Executive Summary
+
+Add a forwarder contract that allows the sequencer client to take multiple actions in the same L1 transaction.
+
+Adjust the sequencer client to batch its actions into a single L1 transaction.
+
+## Introduction
+
+Within the same L1 transaction, one cannot make blob transactions and regular transactions from the same address.
```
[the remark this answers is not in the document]
just-mitch: Yes, sorry, bad typo. Should read "Within the same L1 block..."

in-progress/11103-forwarder-contract/design.md
```diff
@@ -0,0 +1,201 @@
...
+  function forward(address[] calldata _to, bytes[] calldata _data) external onlyOwner {
+    require(_to.length == _data.length);
+    for (uint256 i = 0; i < _to.length; i++) {
+      _to[i].functionCall(_data[i]);
+    }
+  }
+}
+```
+
+Note: this requires all the actions to succeed, so the sender must be sure that, e.g. a failed governance vote will not prevent the L2 block from being proposed.
```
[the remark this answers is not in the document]
just-mitch: I thought about this as well- I landed on that the node should just never send actions that it expects to fail, so I don't think we want it at the moment.

in-progress/11103-forwarder-contract/design.md
just-mitch: Typo. Thank you.

in-progress/11103-forwarder-contract/design.md
just-mitch: Hm I may have been too quick to jump to this. I saw it is in the header today and I didn't see where it is used on L1 so I assumed it was a placeholder without revisiting the forced inclusion design.

I'll strike this section for the time being.

in-progress/11103-forwarder-contract/design.md
just-mitch: `clear` gets a little messy if the onus is on the sequencer I think.

How about instead the `addRequest` has a `validThroughL2Slot` parameter, and when we call `sendRequests` it filters things things that have expired?

in-progress/11103-forwarder-contract/design.md
```diff
@@ -0,0 +1,232 @@
...
+In essence, this class is an API for L1 transactions for the prover node, and a simple wrapper around the `L1TxUtils` class.
+
+### `SequencerPublisher`
+
+The `SequencerClient` will have a `SequencerPublisher` that has many of the same functions currently within the `l1-publisher.ts`, but will have different semantics.
+
+The `SequencerPublisher` will have:
+
+- `requests: RequestWithExpiry[]`
+-
```
LHerskind, line 101: There is an empty dot down here

in-progress/11103-forwarder-contract/design.md
```diff
@@ -0,0 +1,232 @@
...
+
+### Setup
+
+There will be an optional environment variable `sequencer.customForwarderContractAddress` that can be used to specify a custom forwarder contract address.
+
+If this is not set, the sequencer will deploy the Aztec Labs implementation of the forwarder contract, using the Universal Deterministic Deployer, supplying the sequencer's address as the deployment salt, and the sequencer's address as the owner.
+
+### Gas
+
+Including signatures, the calldata for a propose transaction is ~7KB.
```
LHerskind, line 163: Do you recall where the 7kb came from, how many sigs etc? Seems a little high, or maybe the header is bigger than I remember.

in-progress/11103-forwarder-contract/design.md
```diff
@@ -0,0 +1,232 @@
...
+- knowledge of the sequencer's forwarder contract
+
+where:
+
+```typescript
+type Action = "propose" | "claim" | "governance-vote" | "slashing-vote";
+interface RequestWithExpiry {
+  action: Action;
+  request: L1TxRequest;
+  lastValidL2Slot: bigint;
```
LHerskind, line 111: If there could be a comment around the `lastValidL2Slot` that would be neat.

LHerskind (approved):
Some minor comments, but think is good 👍

in-progress/11103-forwarder-contract/design.md
just-mitch: It was actually [your comment](https://github.com/AztecProtocol/engineering-designs/pull/47#discussion_r1916988182). You said `65 * (128 * 2 / 3) + 128 / 3 * 32` for signatures. So probably actually `65 * (48 * 2 / 3) + 48 / 3 * 32` ~= 2.6KB. So plus the other stuff in the header adds about 1KB. So I'll update the math based on 3.6KB.

LHerskind: Alzheimers strikes again.

PR #51, opened 2025-03-10
doc(prd): Fast sync

spalladino:


in-progress/12487-prd-fast-sync.md
```diff
@@ -0,0 +1,57 @@
...
+
+Users SHOULD also be able to specify the location for the snapshot. Protocols supported are ipfs and https. For example:
+`aztec start --node --publisher --sync-method fast --snapshot-url ipfs://...`
+
+Users SHOULD also be able to specify the location for a snapshot index, which is a machine-readable file that lists the latest snapshots. Protocols supported https only. For example:
+`aztec start --node --publisher --sync-method fast --snapshot-index https://...`
+
+### Labs
+
+Snapshots for external networks must be generated and uploaded by Labs to IPFS on at most a 2-week basis, ideally weekly. This process should be automated.
```
iAmMichaelConnor, line 33: Why <= 2 weeks, and ideally weekly?

in-progress/12487-prd-fast-sync.md
```diff
@@ -0,0 +1,57 @@
...
+- Nodes must also be able to specify a snapshot index, and pick the latest one listed.
+- The time for syncing from a snapshot (without validating it) should not depend on the chain length (other than I/O).
+
+### Snapshots index
+
+The format of the snapshots index is left open, but must include the L2 block hash and number, corresponding L1 block number, and either an https or ipfs identifier for obtaining the corresponding snapshot data. This file must be machine-readable, ideally JSON.
+
+### Trust assumptions
+
+For an initial version of this feature, snapshots are trusted, meaning that the downloader trusts the uploader and does not verify its integrity. It only checks that the resulting world state root and latest archive match valid on-chain values.
```
iAmMichaelConnor, line 51: This makes me think to revisit whether we need a "logs tree" (or similar). If I give you a snapshot with a load of purported aztec logs, you have no way to know if they're valid, because they're not committed-to anywhere in the archive tree.

in-progress/12487-prd-fast-sync.md
```diff
@@ -0,0 +1,57 @@
...
+Users SHOULD also be able to specify the location for a snapshot index, which is a machine-readable file that lists the latest snapshots. Protocols supported https only. For example:
+`aztec start --node --publisher --sync-method fast --snapshot-index https://...`
+
+### Labs
+
+Snapshots for external networks must be generated and uploaded by Labs to IPFS on at most a 2-week basis, ideally weekly. This process should be automated.
+
+## Requirements
+
+- Snapshots can be client-specific.
```
iAmMichaelConnor, line 37: Please could you give an example of what a snapshot should / must contain? Surely there are some things that a snapshot _must_ contain?
(Sorry for sending these comments as lots of individual comments - I should've started a review, to reduce noise for you. Each idea came to me one-by-one, after I'd read the whole thing).

in-progress/12487-prd-fast-sync.md
```diff
@@ -0,0 +1,57 @@
...
+### Node operators
+
+Users SHOULD be able to specify which method to use to sync their clients. Methods are either `fast` (snapshots) or `full` (for syncing from the chain). Default is `fast`. For example:
+`aztec start --node --publisher --sync-method fast`
+
+Users SHOULD also be able to specify the location for the snapshot. Protocols supported are ipfs and https. For example:
+`aztec start --node --publisher --sync-method fast --snapshot-url ipfs://...`
+
+Users SHOULD also be able to specify the location for a snapshot index, which is a machine-readable file that lists the latest snapshots. Protocols supported https only. For example:
+`aztec start --node --publisher --sync-method fast --snapshot-index https://...`
```
PhilWindle, line 29: So the idea is the node software would read the index and automatically go and find that snapshot?

spalladino: Yep. In my mind the index is a list of snapshots that gets updated, sorted (sortable) by recency, that nodes query to find out what's the latest one.

in-progress/12487-prd-fast-sync.md
spalladino: Added in a new commit:

> The snapshot should contain all information needed for the client to be considered synced up to a given L2 chain tip. For the current client being built, this is the archiver and world state databases. Snapshot should not include any data whatsoever from the P2P layer. This should only be data syncable from L1. The resulting state of downloading a snapshot should be indistinguishable from a full sync from L1.

in-progress/12487-prd-fast-sync.md
spalladino: How can we validate logs today? Do we have to go to the tx that created them, and look for its log hash? Is there any other way?

in-progress/12487-prd-fast-sync.md
spalladino: A week can be synced in a reasonable amount of time (it's what nodes have been doing with the current chain), and sounded like a nice round number

spalladino: In other words: no specific reason

in-progress/12487-prd-fast-sync.md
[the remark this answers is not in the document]
spalladino: > What further verification steps are needed to fully validate a snapshot? Why aren't those steps considered doable in an initial version?

Exclusively to limit the time required to ship. As for steps, it requires validating both archiver and world state. World state probably requires re-computing all merkle trees, and comparing the resulting root to the one at L1. For archiver, I'm unsure: you probably need to check that the contents correspond to what ended up in world state?

iAmMichaelConnor: Paging @LeilaWang @MirandaWood in case they know the answer :) 

Oh, maybe I remember: The logs are committed-to, in the sense that they become part of a blob, and a blobhash becomes part of an Aztec block header. So actually, maybe my original claim is wrong: we _do_ commit to logs in a verifiable way. I'm not sure if a log-querying node/pxe actually verifies this though: that the log lives in a blob, and the blob hashes to a blobhash, and the blobhash is part of an aztec block header, and the block header is a leaf in the archive tree, and the archive tree root matches an archive tree snapshot.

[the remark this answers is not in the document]
PhilWindle: Oh, I had assumed we were snapshotting the trees. So I just copy down the database files and I have the full trees.

I guess it's a trade-off between the size of the fully populated trees db vs the time taken to re-hash them all. I don't have a n idea of what that trade-off looks like currently.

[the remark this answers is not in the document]
spalladino: > Oh, I had assumed we were snapshotting the trees. So I just copy down the database files and I have the full trees.

Yep, that's the idea for the initial version. But eventually we should validate that what was copied down is sound.

in-progress/12487-prd-fast-sync.md
aminsammara: Blobs gets deleted every 3 weeks, if there's always a snapshot less than 3 weeks old that means for nodes using this sync method, they don't have to rely on a blob sink / super archival full node to. 

1 week felt like a nice round number

in-progress/12487-prd-fast-sync.md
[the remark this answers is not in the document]
iAmMichaelConnor: > But eventually we should validate that what was copied down is sound.

I was going to say: Copying an entire tree doesn't tell you that the root is correct; you _have to_ re-hash all the leaves up to the root, to be convinced that all the upper nodes and root of the tree are correct.

I'd be keen for someone to do it properly, in one fell swoop. It'd mean we don't end up with tech debt, and whoever does it can just immerse themselves in the problem one time. Surely we've got ts library functions already for someone to validate trees and such?

MirandaWood: Yeah - we definitely commit to logs and they can be validated (I think this has always been the case).
We commited to them pre blobs by hashing them with SHA in the rollup circuits, then reconstructing that hash on L1.
Now we publish them (alongside other side effects) in blob(s) on L1. In the rollup circuits we take all the raw side effects, prove they are the fields we are publishing in a blob, and construct an opening proof for each blob (more details [here](https://github.com/AztecProtocol/aztec-packages/blob/master/docs/docs/protocol-specs/data-publication-and-availability/blobs.md)). The opening proofs are published on L1 and verified against blobs sent in the same tx.
If we attempt to publish a different log in the blob vs the real log in the rollups, the blob verification will fail on L1 when publishing. If we attempt to send a different set of blob verification inputs, we would be changing the public inputs to the rollup circuit and the proof verification will fail.
The `blobHash` isn't verified inside the rollup circuits - we don't need to (and it would be expensive SHA). We inject (IIRC) the hash of all `blobHash`es inside the block and use it in calculating the `contentCommitment`. We validate the `contentCommitment` on L1, including getting each `blobHash` from the precompile and checking this hash.
In the pxe (or at least the archiver? the code has changed a lot since my work on blobs) we get blocks by reconstructing them from blobs. So it's not like we start from blocks then validate them later, we build them from fields inside the blob. AFAIK is anything is invalid, the reconstruction will fail. Either way, a user can use the content commitment's `blobHash`es to validate all fields. Or they can use the blob verification inputs from the calldata of the original publish tx.

aminsammara (approved):

PhilWindle (approved):

PR #53, opened 2025-03-12
doc: PRD for Prover Submission Cost

LHerskind:
A PRD for prover submission cost.

### DIFF
--- a/docs/prover-submission-cost/prd.md
+++ b/docs/prover-submission-cost/prd.md
@@ -1,0 +1,134 @@
+# PRD: Prover Submission Cost
+
+- Owner: @LHerskind
+- Approvers:
+  - Product: @joeandrews, @aminsammara
+  - Engineering: @just-mitch, @Maddiaa0
+  - DevRel:
+- Target PRD Approval Date: 2025-03-14
+- Target Delivery Deadline: 2025-03-31
+
+> [!NOTE] Keywords
+> The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119).
+
+# Background
+
+> [!NOTE] Key Terms
+>
+> - **Multi-Proofs**: Scheme that allows redundancy in proof submission and shares payment between all submitters.
+> - **Mana**: The unit of work on the Aztec rollup
+> - **BaseFee**: The amount of fee asset to pay per _Mana_ to cover costs
+> - **Congestion Multiplier**: A multiplier on top of the _BaseFee_ that depends on the usage of the chain to limit spam etc.
+
+The act of producing blocks incurs a cost to the block builders:
+
+- The sequencer pay for publishing the block to the data availability layer and to perform some validity checks such as validating attestations.
+- The prover pays for producing the proof (proving) and then to publish the proof to the base-layer.
+
+We build the background on assuming that you are familiar with the following past design docs:
+
+- Prover Coordination Design([cdfb3a72](https://github.com/AztecProtocol/engineering-designs/blob/cdfb3a72e9b3e4415dcbfe04bd92878996472e6d/in-progress/8509-prover-coordination/design.md))
+- Fee Design Doc ([dac7fdfb](https://github.com/AztecProtocol/engineering-designs/blob/dac7fdfbffb0b0d10ce0dff85221f7a6ece1933b/in-progress/8757-fees/design.md))
+
+The _BaseFee_ was designed to cover:
+
+1. the cost to prove the transactions in the block
+2. the cost to submit the block
+3. `1/N` cost to submit an epoch proof (`N` being epoch size)
+
+The BaseFee calculation originally assumed sufficient mana utilization per block to meet targets. Therefore, blocks with low utilization would not fully cover operational costs through fees alone. To address this, block rewards were introduced as a subsidy.
+
+Provers were compensated with a share of both fees and rewards, allocated according to their provided quotes.
+
+Then the move to [Prover Coordination: Multiproofs](https://hackmd.io/Ivn9axP1SFyEHjpAXVn62g) happened and the costs morphed into the following (`M` is number of proofs):
+
+1. the cost to prove the transactions in the block `M` times
+2. the cost to submit the block
+3. `M/N` cost to submit an epoch proof
+
+However, our model is currently flawed as it is:
+
+- not taking `M` proofs into account when computing the BaseFee
+- paying the proof submission cost to the sequencer, not the prover
+
+These flaws are remnants of the quote-based prover coordination.
+
+## Assumptions
+
+- There exists at least `M` provers that are willing to participate in block production.
+  - If the true number is less than `M` we are overpaying for proofs.
+- The gas estimates of the proof submission cost is somewhat precise
+  - Even though it happens potentially many blocks before the proof submission itself.
+- We expect submission cost to be greater than proving cost.
+
+# User Stories
+
+## End User
+
+As an end user, I want transaction fees to be predictable, transparent, and low, similar to using a single-sequencer system, so that I don’t face unexpectedly high costs.
+
+This is especially important during early network phases with low usage, when high fees might discourage me from continuing to use the network.
+
+> [!WARNING] Multi-proof overhead
+> If the costs associated with submitting multiple proofs (M) are passed directly to users via the BaseFee, it may lead to significantly higher fees compared to centralized or single-proof setups. The additional application of the Congestion Multiplier on these inflated fees could further compound this issue, potentially causing users to abandon the network early on.
+
+## Prover
+
+As an economically rational prover, I wish to participate in block production on the Aztec network, such that I can earn some money on all the machines that I have collected.
+
+In short, as a prover, I wish to:
+
+1. Collect blocks and transactions as the chain grows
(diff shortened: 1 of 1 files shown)


docs/prover-submission-cost/prd.md
```diff
@@ -0,0 +1,138 @@
...
+In short, as a prover, I wish to:
+
+1. Collect blocks and transactions as the chain grows
+2. Run machines to prove transactions and roll them into an epoch proof
+3. Submit epoch proofs to the base-layer for a share of the rewards
+4. Profit???
+
+> [!WARNING]
+> **Multi-proof commentary**   
+> With the nature of the multi-proofs splitting rewards between all submitters and only paying out after. We won't know if there is profit or not until ahead of time.
```
just-mitch, line 90:
```suggestion
> With the nature of the multi-proofs splitting rewards between all submitters and only paying out after. We won't know ahead of time if there is profit or not.
```

docs/prover-submission-cost/prd.md
```diff
@@ -0,0 +1,138 @@
...
+> [!NOTE] 
+> **Key Terms**   
+> - **Multi-Proofs**: Scheme that allows redundancy in proof submission and shares payment between all submitters.
+> - **Mana**: The unit of work on the Aztec rollup
+> - **BaseFee**: The amount of fee asset to pay per _Mana_ to cover costs
+> - **Congestion Multiplier**: A multiplier on top of the _BaseFee_ that depends on the usage of the chain to limit spam etc.
+
+The act of producing blocks incurs a cost to the block builders:
+
+- The sequencer pay for publishing the block to the data availability layer and to perform some validity checks such as validating attestations.
```
just-mitch, line 26:
```suggestion
- The sequencer pays for publishing the block to the data availability layer and to perform some validity checks such as validating attestations.
```

just-mitch (approved):

docs/prover-submission-cost/prd.md
```diff
@@ -0,0 +1,138 @@
...
+
+### Inflation Bounds
+
+- **What**: The inflation of the staking asset **MUST** be less than **X**
+- **Why**: Infinite inflation is not scalable.
+- **Where**: Common sense
+
+> [!NOTE]
+> **HELP**  
+> What should **X** be @aminsammara
```
aminsammara, line 130: If X% is the lifetime maximum allowed supply inflation, it is hard to determine X%. The highest staking rate for similar size networks is 20% (ATOM). It is reasonable to assume that we'll never target more than 80% staked asset percentage. This is about 16% inflation. I'm happy with 20% being the maximum ever allowed protocol supply expansion rate. 

Question: Is it possible to link inflation to the % of staked assets? Therefore inflation is dynamic (within acceptable bounds). The network could benefit from being able to target a desired % staked supply.

docs/prover-submission-cost/prd.md
```diff
@@ -0,0 +1,138 @@
...
+However, our model is currently flawed as it is:
+
+- not taking `M` proofs into account when computing the BaseFee
+- paying the proof submission cost to the sequencer, not the prover
+
+These flaws are remnants of the quote-based prover coordination.
+
+## Assumptions
+
+- There exists at least `M` provers that are willing to participate in block production.
```
aminsammara, line 59: Is it worth adding a requirement that `M` be a governance controlled parameter? Like the block rewards % split between provers and sequencers (correct me please if that's not a governance controlled parameter)

LHerskind: Not really. Who controls the M don't change its existance much, and that assumption was that there exists sufficient provers so don't think much changed there

docs/prover-submission-cost/prd.md
LHerskind: > If X% is the lifetime maximum allowed supply inflation, it is hard to determine X%. The highest staking rate for similar size networks is 20% (ATOM). It is reasonable to assume that we'll never target more than 80% staked asset percentage. This is about 16% inflation. I'm happy with 20% being the maximum ever allowed protocol supply expansion rate.

From this, I am slightly confused, is your answer, X can be 16% or that it should be a maximum staking APR of 20% which will effectively be 16% inflation if there is 80% staked?

I'm gonna assume 16% inflation. 

> Question: Is it possible to link inflation to the % of staked assets? Therefore inflation is dynamic (within acceptable bounds). The network could benefit from being able to target a desired % staked supply.

Yes.

> Do I wanna do it right now?

No. It seems to already be done by how you plan to setup ignition.

aminsammara: To determine max inflation: I looked at some of the highest APRs of similar sized networks (ATOM's 20%), assumed a reasonable upper limit on proportion of assets staked (80%) and arrived at a conclusion that 16% is a reasonable max inflation value. 

Then I bumped it up a few % points because why not and suggested to go with 20%. 

Hope it clears up any confusion.

aminsammara (approved):

docs/prover-submission-cost/prd.md
```diff
@@ -0,0 +1,134 @@
...
+- **Why**: Infinite inflation is not scalable.
+- **Where**: Common economic sends to limit inflation + @aminsammara for number.
+
+# Handling Tradeoffs
+
+When handling the tradeoffs, I believe we should step in the direction of "subsidise too much".
+
+For ignition there will be **no fees** since there are no transactions. If the provers cannot recoup their costs from the block rewards it is very limited who would be able **and** willing to participate.
+
+For alpha, it will also allow us to keep transaction fees low(er) while usage is low, making is easier for users to try the network.
```
Maddiaa0, line 134:
```suggestion
For alpha, it will also allow us to keep transaction fees low(er) while usage is low, making it easier for users to try the network.
```

docs/prover-submission-cost/prd.md
```diff
@@ -0,0 +1,134 @@
...
+> **Key Terms**   
+> - **Multi-Proofs**: Scheme that allows redundancy in proof submission and shares payment between all submitters.
+> - **Mana**: The unit of work on the Aztec rollup
+> - **BaseFee**: The amount of fee asset to pay per _Mana_ to cover costs
+> - **Congestion Multiplier**: A multiplier on top of the _BaseFee_ that depends on the usage of the chain to limit spam etc.
+
+The act of producing blocks incurs a cost to the block builders:
+
+- The sequencer pays for publishing the block to the data availability layer and to perform some validity checks such as validating attestations.
+- The prover pays for producing the proof (proving) and then to publish the proof to the base-layer.
```
joeandrews, line 27: and verify

docs/prover-submission-cost/prd.md
```diff
@@ -0,0 +1,134 @@
...
+In short, as a prover, I wish to:
+
+1. Collect blocks and transactions as the chain grows
+2. Run machines to prove transactions and roll them into an epoch proof
+3. Submit epoch proofs to the base-layer for a share of the rewards
+4. Profit???
+
+> [!WARNING]
+> **Multi-proof commentary**   
+> With the nature of the multi-proofs splitting rewards between all submitters and only paying out after. We won't know ahead of time if there is profit or not.
```
joeandrews, line 90: Is it worth adding in the QOS metric as a requirement here as we discussed? It could be a should not a must, and we can iterate the implementation if we need to mitigate this.

docs/prover-submission-cost/prd.md
```diff
@@ -0,0 +1,134 @@
...
+
+### Scalability
+
+- **What**: In the presence of `<=M` submitters their profit **SHOULD** scale with increasing transaction volume and overall network usage.
+- **Why**: To avoid creating incentives for provers and sequencers to collude, deliberately limiting network throughput to maximize their individual profits at the expense of network growth.
+- **Where**: Derived from concerns regarding potential economic collusion.
+
+### Inflation Bounds
+
+- **What**: The inflation of the staking asset **MUST** be less than 20% yearly
```
joeandrews, line 124: If this is updatable by governance, initial numbers can be much lower, 10% is a fine cap.

docs/prover-submission-cost/prd.md
```diff
@@ -0,0 +1,134 @@
...
+> **Multi-proof commentary**   
+> With the nature of the multi-proofs splitting rewards between all submitters and only paying out after. We won't know ahead of time if there is profit or not.
+
+# Requirements
+
+## Functional Requirements
+
+### Earmarked Fees
+
+- **What**: Fees earmarked for specific actions (e.g., proof submission) **MUST** be directed to the entity incurring the associated cost.
```
joeandrews, line 98: Does this mean the sequencer, or governance dictates who receives the earmarked fee for proving and submission?

joeandrews: i.e is M set by governance?

docs/prover-submission-cost/prd.md
LHerskind: With that argument though it sounds like anything goes. 

If it is set to 10% and it later is updated to be >10% then it does not satisfy the criteria.

The must limit is essentially only there such that if it is from all the other things suggested that you need >20% to make it go round with fees of X, you would still bound it to 20% and need to look at repricing fees instead.

aminsammara: @joeandrews This is a lifetime max cap on annual inflation. It means governance can never set an inflation value above 20% regardless of what's happening.

docs/prover-submission-cost/prd.md
LHerskind: Feel free to suggest the QOS you have in mind.

docs/prover-submission-cost/prd.md
joeandrews: Okay makes sense thanks.

PR #55, opened 2025-03-18
Requirements and Design for Testnet Faucets

just-mitch:
Fix https://github.com/AztecProtocol/aztec-packages/issues/12763
Fix https://github.com/AztecProtocol/aztec-packages/issues/10022

### DIFF
--- a/docs/faucets/dd.md
+++ b/docs/faucets/dd.md
@@ -62,9 +62,9 @@ For users who want to be a validator, they will first:
 
 ```
 curl -s -X POST -H 'Content-Type: application/json' -d \
-'{"jsonrpc":"2.0","method":"node_getBlockHeader","params":[],"id":67}' \
+'{"jsonrpc":"2.0","method":"node_getL2Tips","params":[],"id":67}' \
 "http://localhost:8080" | jq \
-".result.lastArchive.nextAvailableLeafIndex"
+".result.proven.number"
 ```
 
 That will give something like `21314`
@@ -86,6 +86,8 @@ Then, over in discord, they can:
 ```
 
 And this will result in their address getting added to the set.
+
+**Note:** Using the proven chain allows us to add validators even if the pending chain is stalled.
 
 ## Implementation
 
@@ -258,13 +260,13 @@ The `/add-validator` command needs to be augmented to accept a block number and 
 
 When this command is invoked, it should:
 
-1. Ensure the user does not have a "validator" badge already
-2. Query L1 for the specified block number
-3. Ensure that the timestamp associated with the block is from within the last 5 minutes
+1. Ensure the user has not added `X` validators already (using a mapping based on the user's discord ID)
+2. Query L1 for the proven tip of the chain
+3. Ensure that proven tip's block number matches the block number provided by the user
 4. Verify the merkle proof
 5. Mint staking assets for itself
-6. Add the user's address to the validator set, specifying itself as the withdrawer
-7. Apply a new "validator" badge in discord to aid with sybil resistance (see below)
+6. Add the user's address to the validator set, specifying the discord bot's address as the withdrawer
+7. Grant a new "validator" role in discord to the user and bump the user's validator count by 1
 8. Confirm to the user that they were added successfully
 
 #### Risks
@@ -316,7 +318,7 @@ We will need to do a new deployment of the L1 contracts as part of our next upgr
 - [x] FUNC-07: part of the `StakingAssetHandler` interface
 - [x] QUAL-01: anyone can mint fee asset through the `FeeAssetHandler`
 - [/] QUAL-02: we can easily remove anyone added through the `StakingAssetHandler`; all bets are off for validators added otherwise.
-- [x] QUAL-03: requiring a synced node and validator badges in discord should provide a reasonable level of sybil resistance
+- [x] QUAL-03: requiring a synced node and validator role/counts in discord should provide a reasonable level of sybil resistance
 - [x] PERF-01: should be under 24 seconds
 - [x] PERF-02: running the commands should take under 5 minutes. interacting with the bot should be under 24 seconds.
 
@@ -350,7 +352,8 @@ Fill in bullets for each area that will be affected by this change.
 - e2e test that the sequencer client respects the contract-specified mana target
 - Manual testing of adding a validator using the new flow on discord
 - Verify that a validator added via discord may be removed as expected
-- Verify that adding a validator grants a badge which prevents creation of another validator
+- Verify that adding a validator grants a "validator" role in discord
+- Verify that the discord bot enforces a configurable limit of validators per user
 
 ## Documentation Plan
 
--- a/docs/faucets/dd.md
+++ b/docs/faucets/dd.md
@@ -6,7 +6,7 @@
   - @rahul-kothari
   - @LHerskind
   - @signorecello
-- [PRD](https://github.com/AztecProtocol/engineering-designs/blob/b2e744c2ae9f5552e404024edd1a9e20f5e3ea28/docs/faucets/prd.md)
+- [PRD](https://github.com/AztecProtocol/engineering-designs/blob/11ff0c45015027554f13d37f2df85e84090a55a4/docs/faucets/prd.md)
 - Target DD Approval Date: 2025-03-20
 - Target Project Delivery Date: 2025-03-28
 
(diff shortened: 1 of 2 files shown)


docs/faucets/dd.md
```diff
@@ -0,0 +1,363 @@
...
+
+This will be done by manually by the person on the Aztec Labs team managing the discord bot.
+
+### Discord Bot
+
+The `/add-validator` command needs to be augmented to accept a block number and an archive merkle proof.
+
+When this command is invoked, it should:
+
+1. Ensure the user does not have a "validator" badge already
```
aminsammara, line 261: Can we instead specify maximum amount of mints any specific `validator` badge holder can make? Reasoning being is that some teams will want to run multiple validators i.e. 5/10

docs/faucets/dd.md
```diff
@@ -0,0 +1,363 @@
...
+2. Query L1 for the specified block number
+3. Ensure that the timestamp associated with the block is from within the last 5 minutes
```
aminsammara, line 263: Was lasse's comment around stalled block production addressed separately? i.e. in the case block production stopped for more than 5 mins, we cannot add any more new validators.

docs/faucets/dd.md
```diff
@@ -0,0 +1,363 @@
...
+4. Verify the merkle proof
+5. Mint staking assets for itself
+6. Add the user's address to the validator set, specifying itself as the withdrawer
```
aminsammara, line 266:
```suggestion
6. Add the user's address to the validator set, specifying the discord bot's key as the withdrawer
```

docs/faucets/dd.md
```diff
@@ -0,0 +1,363 @@
...
+7. Apply a new "validator" badge in discord to aid with sybil resistance (see below)
```
aminsammara, line 267: wrt comment from above about allowing multiple mints: if approved, this could be changed to "apply new validator badge if they don't have one already"

docs/faucets/dd.md
just-mitch: No, you're right- I'll update it to work off the most recent proven block instead.

docs/faucets/dd.md
```diff
@@ -0,0 +1,363 @@
...
+  - build staking asset handler
+  - tests
+- discord bot dev : 1-2 days
+- run upgrade : <0.5 days
+
+Total: 3-5 days
+
+## Introduction
+
+We are not concerned about the DOS risk of someone holding a large amount of fee asset and flooding the network with transactions: the node and p2p layer should be able to handle this; controlling the production of fee asset needs only to prevent infinite minting (and thus a DOS against the users of the network).
```
LHerskind, line 41: Just a comment to extend here.

We are not concerned, as it can already be done with a minimal amount. Depending on which part you want to dos, you can collect tiny portions of fees and just not spend them and then use them all at once, or you can just create a lot of transactions using the same fees, so seen individually they look fine but most will be reverting.

docs/faucets/dd.md
```diff
@@ -0,0 +1,363 @@
...
+
+This means that we could be maximally minting 7600 / 12 = 633 times per second.
+
+Suppose that we expect the handler to be live for a maximum of 5 years.
+
+Then we have a maximum of 5 \* 365 \* 24 \* 60 \* 60 \* 633 ~= 1e11 mints.
+
+Thus, the maximum mint amount is (2^256 - 1) / 1e11 ~= 1e66.
+
+But we'll err on the side of caution and use 1e42.
```
LHerskind, line 188: Disagree on numbers here. I don't think you would want to ever reach the limit. You don't want to have close to max if you can get away with it, it is just pain for if the token is used in any math. I think we should target a much smaller, that is still enough to cover most everyday things. 1e42 is minting 1e24 or a TRILLION TRILLIONS. I think we can be much more sane with just a 10K or something like that. We just need to evaluate what a few things would cost.

The thing is:
- argue that we can pick a sane number
- and that it will never overflow and cause pain in any meaningful reality

docs/faucets/dd.md
```diff
@@ -0,0 +1,363 @@
...
+
+	constructor(address _owner, address _stakingAsset, address _rollup, uint256 _depositAmount, uint256 _minMintInterval) Ownable(_owner) {
+		STAKING_ASSET = IMintableERC20(_stakingAsset);
+		depositAmount = _depositAmount;
+		rollup = IStakingCore(_rollup);
+		minMintInterval = _minMintInterval;
+	}
+
+	function addValidator(address _attester, address _proposer) external override onlyOwner {
+		require(block.timestamp - lastMintTimestamp >= minMintInterval, "Min mint interval not met");
```
LHerskind, line 228: This one makes sense to me if you are trying to drip people in one at a time, but not if you just want to limit growth daily. I'm not fully getting which is your target here?

If daily, you would mint if out of funds, and "reset" that clock, so you don't need to mint every time. 

Also, this will fail the deposit because it have not approved the rollup to pull funds.

docs/faucets/dd.md
```diff
@@ -0,0 +1,363 @@
...
+	}
+
+	function setMinMintInterval(uint256 _interval) external override onlyOwner {
+		minMintInterval = _interval;
+	}
+
+}
+```
+
+The owner of the handler will be a different EOA, since it will be held by the discord bot.
```
LHerskind, line 249: I would NOT give the owner of this to the bot. Have two roles, one that is owner and can change things, and one that is the bot and can do one thing only: `addValidator`.

docs/faucets/dd.md
```diff
@@ -0,0 +1,363 @@
...
+
+#### Upgrades
+
+The handler will need to be upgraded to the new rollup address when a new rollup is made canonical.
+
+This will be done by manually by the person on the Aztec Labs team managing the discord bot.
```
LHerskind, line 255: Should not be owned by bot.

docs/faucets/dd.md
```diff
@@ -0,0 +1,363 @@
...
+	}
+
+	function setDepositAmount(uint256 _amount) external override onlyOwner {
+		depositAmount = _amount;
+	}
+
+	function setMinMintInterval(uint256 _interval) external override onlyOwner {
+		minMintInterval = _interval;
+	}
+
```
LHerskind, line 245: I would add `setWithdrawer(address _withdrawer)` that can be called by the owner. And not have bot be the owner. Bot should not have admin rights and only minimal access.

docs/faucets/dd.md
LHerskind: As noted earlier, I think it should use a contract specified value and not necessarily the owner. I think that makes more sense as I would not want to have admin rights in the hottest of wallets known to mankind.

docs/faucets/dd.md
```diff
@@ -0,0 +1,366 @@
...
+The `/add-validator` command needs to be augmented to accept a block number and an archive merkle proof.
+
+When this command is invoked, it should:
+
+1. Ensure the user has not added `X` validators already (using a mapping based on the user's discord ID)
+2. Query L1 for the proven tip of the chain
+3. Ensure that proven tip's block number matches the block number provided by the user
+4. Verify the merkle proof
+5. Mint staking assets for itself
+6. Add the user's address to the validator set, specifying the discord bot's address as the withdrawer
```
LHerskind, line 268: As above, I don't think we should have the bot as owner and withdrawer and all. Don't want anything remotely close to admin rights in the hottest of all wallets if we don't need to.

docs/faucets/dd.md
```diff
@@ -0,0 +1,366 @@
...
+
+### Sequencer Client
+
+The sequencer must be updated in its block building logic to ensure that it does not exceed the mana limit for a block as specified on L1.
+
+For now, it can simply query the rollup for `getManaLimit` before building each block; at mainnet, we will be able to move this call to startup, since the rollup will be immutable.
+
+In addition, it must have a new environment variable "MAX_BLOCK_MANA"; when building a block it should stop at the minimum of the contract-specified mana target and the environment variable.
+
+### P2P
```
LHerskind, line 292: Both this and the sequencer client seems fairly separate to the faucets.

LHerskind (commented):
Minor comments on ownership and mint sizes.

docs/faucets/dd.md
```diff
@@ -179,13 +186,21 @@ Then if a block were completely full of `mint` transactions, the maximum number
 
 This means that we could be maximally minting 7600 / 12 = 633 times per second.
 
-Suppose that we expect the handler to be live for a maximum of 5 years.
+Separately, the current amount of mana required for a public transfer is around 1e5 mana.
+
+If we suppose that the base fee is 100 fee asset per mana, then a public transfer will cost 1e7 fee asset.
 
-Then we have a maximum of 5 \* 365 \* 24 \* 60 \* 60 \* 633 ~= 1e11 mints.
+Suppose that a contract deployment costs 100x as much as a public transfer, so 1e9 fee asset.
```
LHerskind, line 193: Don't we have numbers on this? 👀

docs/faucets/dd.md
```diff
@@ -207,28 +222,61 @@ interface IStakingAssetHandler {
...
+
 	uint256 public depositAmount;
 	uint256 public lastMintTimestamp;
 	uint256 public minMintInterval;
+	uint256 public maxDepositsPerMint;
 	IStakingCore public rollup;
+	address public withdrawer;
 
-	constructor(address _owner, address _stakingAsset, address _rollup, uint256 _depositAmount, uint256 _minMintInterval) Ownable(_owner) {
+	modifier onlyCanAddValidator() {
```
LHerskind, line 241: Not something that needs to be updated in this DD. But you can use the `AccessControl` library from open zeppelin do deal with things like this where the are specific roles.

LHerskind (approved):
Thanks for updating 🫡 Minor comments on the design doc, but mainly with a hint for a library that can be used instead of needing to build the access control.

docs/faucets/dd.md
just-mitch: Since our use was so specific I thought to just keep it simpler. Happy to use it though.

docs/faucets/dd.md
```diff
@@ -0,0 +1,418 @@
...
+  - build staking asset handler
+  - tests
+- discord bot dev : 1-2 days
+- run upgrade : <0.5 days
+
+Total: 3-5 days
+
+## Introduction
+
+We are not concerned about the DOS risk of someone holding a large amount of fee asset and flooding the network with transactions: the node and p2p layer should be able to handle this; controlling the production of fee asset needs only to prevent infinite minting (and thus a DOS against the users of the network).
```
PhilWindle, line 41: Is this true though? That we aren't concerned with users minting lots of fee asset and then flooding the network? 

Assuming the node/p2p can cope with it. Doesn't it just do 1 of 2 things?

1. Build up a huge backlog of pointless txs from bots/farmers that hogs the network for a long time.
2. If we set the mempool limits low to avoid a build up, people have their txs evicted. Either way the user exp is terriblke.

docs/faucets/dd.md
just-mitch: I can't find "official" numbers. We don't have Tempo hooked up to masternet, so I can't look at `processTx` traces to grab the gas used for individual transactions. I can see historically though that during network startup, there are transactions that cost ~500K mana. It seems to do a deployment of a schnorr account during this.

aminsammara (approved):

PR #57, opened 2025-03-19
dd: Limiting mempools

spalladino:
Design document for #12879

### DIFF
--- a/docs/mempools/dd.md
+++ b/docs/mempools/dd.md
@@ -64,7 +64,7 @@ To recap, we need to consider:
 - Gas fees and limit vs current base fees and limits
 - Archive root (only on reorgs)
 
-We propose keeping the following indices for all txs:
+We propose keeping the following indices for all txs. These indices are implemented as mappings from the given keys to the tx identifier in the backing LMDB store:
 
 - priority fee
 - fee-payer
@@ -111,8 +111,8 @@ When building a block, for each tx we pick up:
 
 ## Alternative approaches
 
-We can rely heavily on the fact that spamming txs in Aztec is expensive due to the ClientIVC proofs, keep only a global limit on the total number/size of txs, and simply evict based on total mempool size using priority fees. This is much easier to implement than the above.
+We can rely heavily on the fact that spamming txs in Aztec is expensive due to the ClientIVC proofs, keep only a global limit on the total number/size of txs, and simply evict based on total mempool size using priority fees, plus re-checking on every block mined. This is much easier to implement than the above.
 
 An attacker can still spam txs with a shared set of nullifiers to flood the pool with just their txs, but if the priority fee is high enough (if it's too low, the attacker's txs get replaced by other txs), one of those txs will be picked up soon enough and invalidate the others; assuming we filter out the invalid ones fast enough, the sequencer eventually get to other valid txs in the pool. The main assumption is that an attacker cannot produce client proofs at a pace that lets them completely fill the mempool before the next block gets built.
 
-This approach still requires rejecting txs with an ineligible base fee or too large a gas limit, otherwise the attacker could flood the tx with non-executable txs.
+This approach still requires rejecting txs with an ineligible base fee or too large a gas limit, otherwise the attacker could flood the tx with non-executable txs. It also requires reviewing all txs on the mempool whenever a block is mined to drop them based on shared nullifiers, insufficient balance, or max-block-age.
--- a/docs/mempools/dd.md
+++ b/docs/mempools/dd.md
@@ -6,15 +6,16 @@ This addresses how to limit the total number of valid txs. It does **not** addre
 
 ## Tx validity
 
-A tx in Aztec requires the following checks to be valid. Static checks can be done just once, dynamic ones are done when the tx is received and again when the sequencer is building a block. If the tx fails validation during block building, it gets evicted (unless it failed due to _Gas fees over block base fee_).
+A tx in Aztec requires the following checks to be valid. Static checks can be done just once, dynamic ones are done when the tx is received and again when the sequencer is building a block. If the tx fails validation during block building, it gets evicted, unless it failed due to gas fees lower than the block base fee.
 
 ### Static
 
-- Chain id
+- L1 chain id and L2 version
 - Setup function
 - Correct public execution requests and logs
 - Minimum gas fees (base and priority)
 - Reasonable gas limit (currently missing)
+- Valid ClientIVC proof
 
 ### Dynamic
 
@@ -23,7 +24,7 @@ A tx in Aztec requires the following checks to be valid. Static checks can be do
 - Archive root exists (can become invalid after a reorg)
 - Fee juice balance for fee payer
 - Gas fees over current block base fee
-- Gas limit below current block target
+- Gas limit below current block limit
 
 ## Geth
 
@@ -67,16 +68,28 @@ We propose keeping the following indices for all txs:
 
 - priority fee
 - fee-payer
-- nullifier
+- nullifiers (indexes a tx by all of its nullifiers)
 - base fee
 - gas limit
 - max block number
 
-When adding a tx, in addition to running all static checks:
+When adding a tx, we first run the trivial checks:
+
+- Correct L1 chain id and L2 version
+- Public setup function is acceptable
+- Correct public execution requests and logs
+- Gas fees (base and priority) above a given minimum
+- Valid ClientIVC proof
+- Max block number for inclusion is in the future
+- Double spend (repeated nullifiers) against existing state
+- Gas limit is below the current block gas limit
+- Archive root exists
+
+And then:
 
 - We check if the current balance of the fee payer, minus the max cost of all pending txs for that fee payer, is enough to pay for this tx. If it is not, we try evicting other txs with a lower priority fee. If that works, and all other checks pass, we include the tx dropping the others.
-- We check if it shares a nullifier with any existing pending tx. If it pays more than all of the conflicting ones, and it passes all other checks, we include it and drop the other ones.
-- We check if the tx gas limit is below the current block gas limit, and its fees are above the current base fees. If not, we drop it. Note that we could save it for later in case fees drop in the future, but this means tracking two different pools (executable and non-executable, as geth does).
+- We check if it shares a nullifier with any existing pending tx (we already checked duplicates against current state at this point). If it pays more than all of the conflicting ones, and it passes all other checks, we include it and drop the other ones.
+- We check if the tx fees are above the current base fees. If not, we drop it. Note that we could save it for later in case fees drop in the future, but this means tracking two different pools (executable and non-executable, as geth does).
 - We check if we are below a configurable size/number of pending txs. If we are not, start dropping txs with lower priority fee (sorted by priority fee) until we get again below the threshold.
 - If we do add the tx, we index its max block number as the minimum of the tx's max-block-number and the current block number plus a configurable number. This allows us to evict txs after they'd been sitting in the pool for a very long time.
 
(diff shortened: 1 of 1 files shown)


docs/mempools/dd.md
```diff
@@ -0,0 +1,106 @@
...
+
+Our client tx mempool today grows unbounded. We need to place a limit on how many txs it absorbs.
+
+This addresses how to limit the total number of valid txs. It does **not** address how to protect against DOS attacks that cause the client to validate more txs than it can process.
+
+## Tx validity
+
+A tx in Aztec requires the following checks to be valid. Static checks can be done just once, dynamic ones are done when the tx is received and again when the sequencer is building a block. If the tx fails validation during block building, it gets evicted (unless it failed due to _Gas fees over block base fee_).
+
+### Static
```
LHerskind, line 11: Should static also be a valid proof for the private chunk?

docs/mempools/dd.md
```diff
@@ -0,0 +1,106 @@
...
+- Reasonable gas limit (currently missing)
+
+### Dynamic
+
+- Max block number for inclusion
+- Double spend (repeated nullifiers)
+- Archive root exists (can become invalid after a reorg)
+- Fee juice balance for fee payer
+- Gas fees over current block base fee
+- Gas limit below current block target
```
LHerskind, line 26: We are having a target and limit similarly to 1559, so I think we would allow as long as it below block limit and not "just" the target.

docs/mempools/dd.md
```diff
@@ -0,0 +1,106 @@
...
+- Conflicting nullifiers
+- Max block number
+- Gas fees and limit vs current base fees and limits
+- Archive root (only on reorgs)
+
+We propose keeping the following indices for all txs:
+
+- priority fee
+- fee-payer
+- nullifier
```
LHerskind, line 70: All the tx nullifiers or just the "tx-hash"? (if that is still done)

docs/mempools/dd.md
```diff
@@ -0,0 +1,106 @@
...
+- base fee
+- gas limit
+- max block number
+
+When adding a tx, in addition to running all static checks:
+
+- We check if the current balance of the fee payer, minus the max cost of all pending txs for that fee payer, is enough to pay for this tx. If it is not, we try evicting other txs with a lower priority fee. If that works, and all other checks pass, we include the tx dropping the others.
+- We check if it shares a nullifier with any existing pending tx. If it pays more than all of the conflicting ones, and it passes all other checks, we include it and drop the other ones.
+- We check if the tx gas limit is below the current block gas limit, and its fees are above the current base fees. If not, we drop it. Note that we could save it for later in case fees drop in the future, but this means tracking two different pools (executable and non-executable, as geth does).
```
LHerskind, line 79: If the gas limit is above the current block gas limit would we not always just drop it since it "might" be too big to include. At least for until that value is updated (which I assume is why you asked static or dynamic earlier? 👀)

docs/mempools/dd.md
```diff
@@ -0,0 +1,106 @@
...
+- We check if we are below a configurable size/number of pending txs. If we are not, start dropping txs with lower priority fee (sorted by priority fee) until we get again below the threshold.
+- If we do add the tx, we index its max block number as the minimum of the tx's max-block-number and the current block number plus a configurable number. This allows us to evict txs after they'd been sitting in the pool for a very long time.
```
LHerskind, line 81: Maybe i missed it explicitly, but also if the max block number is in the past we just drop it already. The max block number was mentioned earlier for dynamic, but not explicitly discarding etc. Maybe just me being braindead because late.

docs/mempools/dd.md
```diff
@@ -0,0 +1,106 @@
...
+- fee-payer
+- nullifier
+- base fee
+- gas limit
+- max block number
+
+When adding a tx, in addition to running all static checks:
+
+- We check if the current balance of the fee payer, minus the max cost of all pending txs for that fee payer, is enough to pay for this tx. If it is not, we try evicting other txs with a lower priority fee. If that works, and all other checks pass, we include the tx dropping the others.
+- We check if it shares a nullifier with any existing pending tx. If it pays more than all of the conflicting ones, and it passes all other checks, we include it and drop the other ones.
```
LHerskind, line 78: Or nullifier already in the state ye? Earlier alluded to duplicates, just adding explicit comment here.

docs/mempools/dd.md
```diff
@@ -0,0 +1,106 @@
...
+- We check if the tx gas limit is below the current block gas limit, and its fees are above the current base fees. If not, we drop it. Note that we could save it for later in case fees drop in the future, but this means tracking two different pools (executable and non-executable, as geth does).
+- We check if we are below a configurable size/number of pending txs. If we are not, start dropping txs with lower priority fee (sorted by priority fee) until we get again below the threshold.
+- If we do add the tx, we index its max block number as the minimum of the tx's max-block-number and the current block number plus a configurable number. This allows us to evict txs after they'd been sitting in the pool for a very long time.
+
+When a new block is mined:
+
+- We drop all txs that share nullifiers with nullifiers from the mined blocks
+- We update the balance of fee payers and drop txs that can no longer be paid
+- We drop all txs with a computed max-block-number equal or lower than the mined one
+- We drop all txs that no longer fit in a block due to gas limit, and that have a base fee lower than the new block's
```
LHerskind, line 88: I don't fully follow the gas limit part here? We have dynamic, but not in the same ethereum sense where it is block producers directly influencing it. At least not in the current setup.

docs/mempools/dd.md
```diff
@@ -0,0 +1,106 @@
...
+  - Note that, if we check duplicates against existing nullifiers on every block we add, we only need to check against nullifiers emitted during the block being built.
+- We check gas fees and limits against the current block base gas fees and limits. If we fail, we just skip the tx.
+
+## Alternative approaches
+
+We can rely heavily on the fact that spamming txs in Aztec is expensive due to the ClientIVC proofs, keep only a global limit on the total number/size of txs, and simply evict based on total mempool size using priority fees. This is much easier to implement than the above.
+
+An attacker can still spam txs with a shared set of nullifiers to flood the pool with just their txs, but if the priority fee is high enough (if it's too low, the attacker's txs get replaced by other txs), one of those txs will be picked up soon enough and invalidate the others; assuming we filter out the invalid ones fast enough, the sequencer eventually get to other valid txs in the pool. The main assumption is that an attacker cannot produce client proofs at a pace that lets them completely fill the mempool before the next block gets built.
```
LHerskind, line 104: I guess this would in practice depend on how big blocks we will be able to construct, e.g., if the blocks are teeny tiny, it would be simpler to pull off no? Or possible to "slowly but faster than block production" grow your share of the mem-pool until filled.

docs/mempools/dd.md
```diff
@@ -0,0 +1,106 @@
+# Limiting Client Mempools Design Document
+
+Our client tx mempool today grows unbounded. We need to place a limit on how many txs it absorbs.
+
+This addresses how to limit the total number of valid txs. It does **not** address how to protect against DOS attacks that cause the client to validate more txs than it can process.
+
+## Tx validity
+
+A tx in Aztec requires the following checks to be valid. Static checks can be done just once, dynamic ones are done when the tx is received and again when the sequencer is building a block. If the tx fails validation during block building, it gets evicted (unless it failed due to _Gas fees over block base fee_).
```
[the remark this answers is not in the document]
spalladino: Rewrote as:

> If the tx fails validation during block building, it gets evicted, unless it failed due to gas fees lower than the block base fee.

docs/mempools/dd.md
spalladino: Got it. Dropping it.

docs/mempools/dd.md
spalladino: Not sure if block size affects it, I think it's more related to block frequency? The goal of the attacker is flood the mempool at minimal cost for them, so they want the mempool to contain only their txs, but such that only one of those is executable. As soon as a block is mined that includes one of those txs, all of the other txs from the attacker are discarded, and they need to re-flood. I may be missing something though.

docs/mempools/dd.md
```diff
@@ -0,0 +1,118 @@
...
+  - [If the tx pool is full](https://github.com/ethereum/go-ethereum/blob/80b8d7a13c20254a9cfb9f7cbca1ab00aa6a3b50/core/txpool/legacypool/legacypool.go#L691-L692), it discards cheaper txs based on gas tip (ie priority fees). Only the global slots seem to be considered here.
+- When cleaning up the pool in a loop, loops over pending (executable?) txs for every sender that has gone over AccountSlots, and drops txs (based on nonce) form them. Also loops over future (non-executable?) txs based on **heartbeats**: accounts with the most time without any activity get their txs pruned first.
+
+## Difficulties
+
+In addition to all complications that Ethereum has, we also have the issue that a tx public execution can invalidate an arbitrary number of existing txs just by emitting nullifiers. We have no way of knowing that in advance.
+
+Also, while for Ethereum a "replacement" is just a tx with the same nonce and sender as an existing one, for us any tx that shares a nullifier can technically be a replacement. This also means that tx A may be a replacement for B and C, but B and C may be unrelated to each other.
```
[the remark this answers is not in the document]
spalladino: Yes, that's what I had in mind. Solving the optimal case for which txs to drop and which ones to keep to maximize tips is not trivial, and a greedy algorithm that always prioritizes highest paying tx first should be good enough.

docs/mempools/dd.md
```diff
@@ -0,0 +1,118 @@
...
+
+When a new block is mined:
+
+- We drop all txs that share nullifiers with nullifiers from the mined blocks
+- We update the balance of fee payers and drop txs that can no longer be paid
+- We drop all txs with a computed max-block-number equal or lower than the mined one
+
+Note that we should not be dropping them, but rather pushing them to the side to reincorporate them in case of a reorg. But we will dismiss this for now.
+
+When a reorg happens, we crawl through all txs and evict the ones with a no-longer-valid archive root. We could also do this via an index, depending on how frequent we think reorgs will be.
```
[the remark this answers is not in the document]
spalladino: We still need to support reorgs. What I meant is that, in the following scenario:
- A block B1 is mined
- In that block, there's a tx TX1 with a nullifier N1
- We have another tx TX2 with nullifier N1 in our mempool
- We evict that tx from our mempool
- A bit later, there's a reorg, that pushes B1 out of the chain
- Then TX2 is valid again, and we could bring it back to the mempool

The last step is out of scope for now. We're fine _not_ bringing TX2 back.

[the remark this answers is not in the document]
just-mitch: When this DD goes in, and issues get created for it, can we also create an issue for a followup design to tackle this reorg scenario? 

That would just be such a terrible UX.

docs/mempools/dd.md
just-mitch: I don't see what about the current design prevents this same scenario?

docs/mempools/dd.md
```diff
@@ -0,0 +1,118 @@
...
+
+To recap, we need to consider:
+
+- Balance of fee-payers
+- Conflicting nullifiers
+- Max block number
+- Gas fees and limit vs current base fees and limits
+- Archive root (only on reorgs)
+
+We propose keeping the following indices for all txs:
```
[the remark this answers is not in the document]
spalladino: They are just LMDB mappings. I've just clarified it!

docs/mempools/dd.md
just-mitch: So really the proposed design is identical to this alternative design, its just that the maintenance of the indices should give much better performance when you have to rescan for invalid transactions (which you must do after every transaction you process, and every every block that is mined)?

spalladino: The main difference is the criteria for accepting new txs. In the simplified design, you validate the tx against the current chain state upon receiving it, and if the mempool is full, drop a tx with lower priority fee to make room for it. In the full design, you validate the tx against not just the current state but also what's on the mempool.

just-mitch (approved):

docs/mempools/dd.md
just-mitch: Got it. Thanks!

natebeauregard (approved):

PhilWindle (approved):
A complicated subject that's for sure. Thanks for tackling this.

PR #63, opened 2025-05-20
Initial design/thoughts around tx availability

PhilWindle:


in-progress/13000-tx-availability/design.md
```diff
@@ -0,0 +1,91 @@
...
+
+1. Proposers are assumed to already have access to the transactions they are including in the block.
+2. Validators need to have access to the transactions to re-execute and attest to the block. This needs to be achieved in a short period of time, validators have in the order of 10 seconds to attest to a block.
+3. Provers also require transactions for re-executing and proving the epoch. This is less time critical as proving is performed over the course of 2 epochs.
+4. Additionally, both validators and provers need to verify the proofs that accompany the transactions, these proofs dominate the size of the transaction payload.
+5. From a network perspective, we are satisfied with 66% of validators successfully re-executing and attesting to a block. The network only requires 1 prover to submit a proof of an epoch.
+
+When it comes to hardware and bandwidth requirements, it is desirable for people to be able to operate as validators with a single consumer grade machine and home broadband connectivity with potentially limited upload bandwidth. Provers can be expected to have significantly more resources, these are likely to be professional/institutional organisations.
```
just-mitch, line 29: It would be good to specify the exact minimum supported up/down speeds.

in-progress/13000-tx-availability/design.md
```diff
@@ -0,0 +1,91 @@
...
+
+1. Request/Response. Validators and provers will randomly select subsets of their total peers and directly request the transactions that they are missing. This is done repeatedly over a period of rounds until the transaction is found or an arbitrary time limit is reached. Nodes are careful when using this so as to not be too aggressive in their requests. Nodes employ peer-scoring techniques to rate limit peers that are requesting too much data. As a result, request-response is not a completely reliable mechanism for transaction retrieval.
+2. Prover Coordination. Provers can be configured with any number of http urls that they can use to retrieve transactions. This enables provers to run a number of additional nodes on the network for the purpose of increasing their chances of receiving everything they require.
+
+## Proposed Solution
+
+Gossiping is already very effective at transaction propagation and it is assumed that with further reductions in proof size/increases in bandwidth availability this will continue to be the case. However, due to the reasons outlined above, we have to assume that there will be instances of nodes not having the transactions they require.
```
just-mitch, line 48: How "effective"?

in-progress/13000-tx-availability/design.md
```diff
@@ -0,0 +1,91 @@
...
+
+As a leftover from the previous method of prover coordination, provers have additional configuration allowing them to request transactions from other nodes via http. This can be made a universal option for all nodes and incorporated into the TxRetrieval workflow.
+
+### Increasing prover transaction pool requirements
+
+Provers are assumed to be organisations with access to considerable hardware resources and bandwidth capabilities. It is not unreasonable for them to:
+
+1. Run multiple full nodes in addition to the prover node.
+2. Configure every node with a very large transaction pool size, significantly reducing the transaction eviction rate.
+3. Configure every node to have a high peer count, increasing the likelihood of request/response success.
```
mralj, line 87: We could increase the mesh size to D = 12 to be more aggressive. 
AFAIK the values above have no observable improvements

in-progress/13000-tx-availability/design.md
```diff
@@ -0,0 +1,91 @@
...
+
+Therefore, we propose a solution that adds additional layers/protocols and enhancements to the current gossipsub + request/response methods encapsulated with a new TxRetrieval module. TxRetrieval will continuously work to retrieve all transactions specified as part of block proposals until either the transaction has been retrieved or the transaction is not included in a mined block. This latter case is identifed by the publishing of a block that does contain the transaction and where the transaction has not been included in a prior mined block.
+
+### Request/Response from proposal propagater
+
+Upon receipt of a block proposal where the receiving peer P does not have all required transactions, P will start a request reponse process but will more aggresively target the peer propagating the block proposal. That is to say, that each round of peer sampling will always include the propagating peer. The block proposal will be propagated further immediately, regardless of whether P is missing transactions or not. This ensures that the proposal is not delayed.
```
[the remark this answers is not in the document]
mralj: > Why wouldn't we just have the proposer send the proposal directly to the validators?

What if the proposer is not connected via P2P to he validators? I don't think it would be easy to guarantee this.

in-progress/13000-tx-availability/design.md
mralj: From this doc: https://research.protocol.ai/publications/gossipsub-v1.1-evaluation-report/vyzovitis2020.pdf

> The latency distribution shows that most nodes receive messages under the 150 ms threshold in gs-v1.1
with p99 at 165 ms, while in gs-v1.0 the threshold is 200 ms with p99 at 192 ms. This is because

The setup was 1k nodes, 100 nodes are active publishers, rest 900 hundred are passive

in-progress/13000-tx-availability/design.md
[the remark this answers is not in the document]
just-mitch: Committees have an entire aztec epoch (30 minutes or so) to prepare for their active epoch. That would be plenty of time to troll through the network and form a *decent* mesh.

Though yes, we couldn't rely on it entirely: it would be best effort and some validators would still receive proposals on the slow path.

Regardless, if propagation time is truly ~200ms at p99 regardless, then this is moot.

in-progress/13000-tx-availability/design.md
just-mitch: Thanks! Looks like the messages in the test were 2KB. @PhilWindle How big are out proposals?

in-progress/13000-tx-availability/design.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Proposed Solution
+
+Gossiping is already very effective at transaction propagation and it is assumed that with further reductions in proof size/increases in bandwidth availability this will continue to be the case. However, due to the reasons outlined above, we have to assume that there will be instances of nodes not having the transactions they require.
+
+The ultimate solution would be to do as Ethereum does and bundle the transactions with the block proposal. This would ensure that everyone who requires the transactions would have them. However, this is not feasible. Transactions are approximately 60KB compressed. A block proposal containing transactions at a rate of 10 TPS will contain 360 transactions = ~22MB of data. 
+
+Our usage of Gossipsub works such that:
+
+1. The original publisher of a message sends to all peers. This number is configurable but likely to be no less than 50.
```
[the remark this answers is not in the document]
PhilWindle: We currently have this switched on by default. It's configuration on the node so can easily be switched off. Do you see any downside to having it on? It seems like quite a good feature to kickstart propagation around the network.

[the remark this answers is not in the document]
mralj: > We currently have this switched on by default.

I completely missed this 🤦 

>  Do you see any downside to having it on? 

Yes. If we implement a solution where peer P' asks for the missing transactions from peer P (proposal sender), we could _overburden_ it by ending up in a situation where a bunch (up to `maxPeerCount`, which is currently 100) are asking the proposer for the missing txs simultaneously. 
Furthermore, some of these peers could be physically far away from the Proposer, so depending on the network, we could either waste bandwidth because they'll be receiving the proposal from other peers(which I don't think is a huge deal) or, worse, they could receive the proposal from the Proposer and ask it the transactions effectively loosing latency. 

I feel relying on the mesh is a better choice, and I'd like to ask the validators to increase their mesh count (`D`) to 12 because it will give us better spread w/o losing much (at all?) latency.

Still, this is just my feeling/opinion/reasoning; I don't have any concrete numbers, experiments, or proofs. In practice, either option shouldn't make a huge difference.

PR #65, opened 2025-06-10
feat: staking entry queue dd

just-mitch:
Staking entry queue with the ability to build up a large backlog, then flush them all, then go into "normal" operation.


docs/staking-entry-queue/dd.md
```diff
@@ -0,0 +1,136 @@
...
+  function flushEntryQueue() external;
+```
+
+## Implementation
+
+When would-be validators call `deposit`, they will enter a queue maintained by the rollup instance.
+
+Validators in the queue do not participate in block production, and will not be sampled during committee selection.
+
+When _anyone_ calls `flushEntryQueue` for a given epoch `e`, then up to `N` validators will be dequeued and added to the validator set, and be eligible to serve on the committee for epoch `e`.
```
LHerskind, line 68: I think the notion of "being able to ser ve on `e` is not good here. For the simple reason that it depends on the validator sampling, so if the "delay" in there changes I don't see why it should change much in the queues. It is fine that it is only once per epoch so there is some time frame where it can happen, but would not say anything about the time where they are eligible as that is really depending on the other parts, mainly that when queue is flushed they get added to the validator set

docs/staking-entry-queue/dd.md
```diff
@@ -0,0 +1,136 @@
...
+
+_NOTE_: we will restrict it such that `flushEntryQueue` may only be called once for epoch `e`, and only during epoch `e-1`.
```
LHerskind, line 70: It don't make sense to me to say that it may be called once for epoch `e` and only during `e-1` when there are no inputs. So think it can be said simpler, just that it can be called at most once every epoch 🤷 Or you can see the amount of validators that can be flushed as a curve that accumulates or something

docs/staking-entry-queue/dd.md
```diff
@@ -0,0 +1,136 @@
...
+
+_NOTE_: we currently setup epochs 1 full epoch in advance, so in the first slot in epoch `e-1` we can setup epoch `e`. So in most cases, it will be that the queue is drained during epoch `e-1`, and the new validators may appear in the committee for epoch `e`, and that will be know effectively at the outset of epoch `e-1`. If no one is calling `setupEpoch` though for multiple epochs, there could be a build up of new validators that could be eligible, all in the same epoch.
```
LHerskind, line 72: As mentioned earlier, they won't be in the committee for `e` because that is controlled separately, but they will be in the validator set. Think I would try to avoid all of these `e-1` and `e` notions

docs/staking-entry-queue/dd.md
```diff
@@ -0,0 +1,136 @@
...
+
+### Selection of `N`
+
+See the notebook for details.
+
+```
+cd notebook
+uv sync
+uv run marimo edit sample.py
```
LHerskind, line 81: I would suggest not hiding fields in the `sample.py` unless just big impls for say queue etc, but keep it and then rely on the app-view if people are to not look at it. e.g., you can do `marimo run sample.py` and then you will essentially get the view where all the ones you are hiding are hidden.

docs/staking-entry-queue/notebook/sample.py
```diff
@@ -0,0 +1,161 @@
...
+    import math
+    import matplotlib.pyplot as plt
+    import numpy as np
+    from scipy.stats import hypergeom
+    return hypergeom, math, mo, plt
+
+
+@app.cell(hide_code=True)
+def _(mo):
+    existing_validators = mo.ui.slider(
```
LHerskind, line 19: You can do something like `show_value=True` which is quite nice for seeing the value without needing to hover. If you make the full width as well also simpler to hit certain values.

Alternative approach:

```python
existing_validators = mo.ui.slider(
    100,
    8192,
    value=4096,
    label="Total Existing Validators",
    show_value=True,
    full_width=True,
)
new_validators_added = mo.ui.slider(
    1,
    256,
    value=204,
    label="New Validators Added to Pool",
    show_value=True,
    full_width=True,
)
committee_size = mo.ui.slider(
    32, 64, value=48, label="Committee Size", show_value=True, full_width=True
)

mo.vstack([existing_validators, new_validators_added, committee_size])
```

docs/staking-entry-queue/notebook/sample.py
```diff
@@ -0,0 +1,161 @@
...
+            new_validators_added.value,     # n: Successes in pool
+            committee_size.value           # N: Sample size
+        )
+
+    epoch_duration_s = 32*36
+    num_drains_for_double = math.ceil(committee_size.value / new_validators_added.value)
+    seconds_per_day = 60*60*24
+
+    result_text = f"""
+    A committee of **{committee_size.value}** members requires **{k_threshold}** seats for ⅓ control.
```
LHerskind, line 142: I don't agree with the implicit premise that it is only important to protect the next epoch from 1/3 control. Namely, the more important thing to protect against is generally fast takeovers. For example, while I might not take over the next committee, with 1/20, and there being 75 epochs a day, I can become a supermajority of the set (almost 80%) within just a single day. I see that as a massive problem. Essentially, if I have enough money, I can take over the network in less than 24 hours - even less if I turn the slasher against the network and slash other validators as soon as I have a majority.

I think beyond the probabilities of taking over next, we should take the approach that you should not be able to "become emporer" faster than "x" time. Ensuring that if the "takeover" is noticed, there are actualy time to react to it.

The implementation stays the same as this can just be modelled as a different value for `num_existing_validators // N` hence the time can be converted into speed.

LHerskind (commented):
Generally I think it makes sense, but I would alter the premise of churn limits.

PR #69, opened 2025-06-30
dd: Optimistic attestation signature validation

spalladino:


docs/optimistic-attestation-signatures/dd.md
```diff
@@ -0,0 +1,168 @@
...
+
+Considering the above, what should an L2 node do if they see two blocks N and N+1 in L1, both from the same epoch, where N does not have its attestations? While this situation should not happen since the proposer for N+1 should refuse to build on N, there is nothing in the rollup contract that prevents it from happening. And since attestation for a block is economically equivalent to an attestation to also all its previous blocks within the same epoch, L2 nodes could happily accept both blocks N and N+1 in this example.
+
+It's unclear to me whether this may lead to situations where proposers purposefully omit attestations for a block, knowing that this gets "patched" in the following one. This doesn't seem to be the case if the attestation committee refuses to sign off N+1 given the lack of attestations on L1 for N, but I still wanted to flag it.
+
+The open question remains on whether L2 nodes should accept blocks N and N+1 in the example above, or wait until their epoch gets proven. For simplicity, I'd push for only accepting such blocks once they get proven.
```
aminsammara, line 99: Rephrasing the discussion here as "should proposers build on top of (potentially valid) blocks that have no attestations?"

This is equivalent to -  borrowing your terminology - taking off the training wheels provided by the attestations. If we don't trust our proving system, wouldn't we then impose the requirement that attestations are posted? If we don't, then the equilibria is that no one posts any attestations except for the last proposer who must post them (or just make them available to the prover - depending on design). 

If this is an acceptable outcome then probably better to remove attestations altogether.

spalladino: > Rephrasing the discussion here as "should proposers build on top of (potentially valid) blocks that have no attestations?"

I agree that the answer to that discussion should be "no". But we still need to consider what should a node do if, for whatever reason, this happens (since it is a valid state in which the system can be).

spalladino: @The-CTra1n I assume you're referring here to moving attestations to blobs, which is a separate discussion? Moving stuff to blobs increases complexity, yes, not just for providing the KZG proof but also because it changes our blob layout, which impacts other circuits. Still, if we decided to go down that route, I think that the KZG proof is easier than collecting votes.

spalladino: > You have better line of sight on the circuit changes necessary though

I don't, I'll have to bring this up with @iAmMichaelConnor heh

docs/optimistic-attestation-signatures/dd.md
```diff
@@ -0,0 +1,166 @@
...
+  let digest = block.proposalDigest
+  require(ecrecover(attestations[invalidIndex], digest) !== committee[invalidIndex])
+  storage.tips = storage.tips.updatePendingBlockNumber(blockNumber - 1)
+```
+
+Considering we need to do only a single `ECRECOVER`, we can estimate the gas for this operation to be `3000 * 31 = 93k` gas less than the current proposal validation, plus `4200` for the two SLOAD operations (`attestationsHash` and `committeeCommitment`), and the 38400 gas for calldata. This results in `160k - 93k + 4.2k + 38k = 110k` gas. Note that this function should hardly ever be called.
+
+### `submitProof`
+
+In addition to verifying the rollup validity proof, `submitProof` also needs to check the validity of attestations in the last block in the epoch.
```
iAmMichaelConnor, line 128: Why?

docs/optimistic-attestation-signatures/dd.md
```diff
@@ -0,0 +1,166 @@
...
+### Where do attestations get posted?
+
+While the cheapest option is not to post attestations at all, and require these to be sourced from the p2p layer in order to invalidate a block, we quickly run into data availability issues, since the invalidator may not have access to the invalid attestations in order to post them to L1 to trigger the invalidation.
+
+Two options remain: posting them to calldata or to blobs. The flow in both cases is similar: proposers post attestations in either of them, and store in L1 a commitment to them (we can also modify the block hash to include a commitment to a set of attestations, to avoid an extra `SSTORE`, but this is a larger change). On block proposal, we check that the hash corresponds to the data posted. On (in)validation, the caller re-posts the attestations to L1, which get re-hashed and compared against the stored commitment, and then verified.
+
+Calldata for a 48-sized committee is `(48 * 2/3 * 65) + (48 * 1/3 * 20) = 2400` bytes, or `38400` gas (note that after [EIP7623](https://eips.ethereum.org/EIPS/eip-7623) this could shoot up to `96k` depending on execution gas). This can be saved in favor of moving the attestations to blocks.
```
just-mitch, line 73: That EIP7623 went live in pectra, so we're living in a 96K world now. 

Also, now that we have [EIP2537](https://eips.ethereum.org/EIPS/eip-2537), we might should revisit BLS. @LHerskind  had calculated that it was quite more expensive to use BLS *with verification* at a committee of 48, but if in the happy path, all we pay for is ~100B calldata and storing ~4 words (which can be in a roundabout so basically free), plus some new complexity in validator registration and setting up an epoch, it could be worth it.

spalladino: That's really really interesting

docs/optimistic-attestation-signatures/dd.md
```diff
@@ -0,0 +1,166 @@
...
+
+Since invalidating a block is an expensive operation, a proposer can be given a gas rebate for this action. This rebate could be taken from the proposer who posted the block with invalid attestations. However, this means enshrining a slashing mechanism in the system, which involves additional complexity. We could initially require proposers (or provers) to absorb this additional expense, assuming rewards will be enough to offset it.
+
+Alternatively, we can keep this method open for anyone to call. Assuming there is no reward for this action, we expect only block proposers to actually call it. Should we introduce a gas rebate and reward, we could end up with multiple nodes racing to claim this reward.
+
+Given the incentives, I suggest keeping the method open and with no rewards. And to minimize complexity, I suggest no gas rebates at all.
```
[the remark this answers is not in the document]
The-CTra1n: Eventually, the offending proposer needs to be slashed to cover the cost of the honest proposer trying to fork out the offending block: this is an objective (slashable) offense. To start though, tuning block rewards to cover any need to fork a bad block should be fine

PR #73, opened 2025-07-11
dd: Circuit changes for building in chunks

spalladino:
Adds a spec for the circuit and protocol changes needed for building in chunks with chained txs.


docs/building-in-chunks/circuits.md
```diff
@@ -0,0 +1,178 @@
...
+
+## Fully decoupling L2 blocks from L1
+
+Should we ever want to decouple L2 block production from L1 checkpointing, there are a few changes needed to the design above. For clarity, by decoupling L2 blocks from L1 checkpointing, we mean having L2 produce blocks, and at some point an entity groups some of them into a checkpoint and pushes them to L1, but L2 can keep advancing independent of L1. In other words, L2 liveness should not depend on L1 availability (though its finality most definitely should).
+
+Since we do not know _when_ a chunk will make it to L1, we cannot know in which blob it will land, so we cannot know its `blobSponge` start and end state. One option here would be to just keep a rolling hash of `TxEffects` which gets permanently accumulated and is never reset, so when the network decides to aggregate a few chunks into a checkpoint, the blob verification circuits need to account for a non-zero initial sponge (I need to check feasibility of this).
```
iAmMichaelConnor, line 170: Very clever idea.
This should be possible. At the moment we assert that the "start" sponge of a block is empty. Instead, we can assert that the start sponge of a chunk is always that of the previous -- even for the 1st chunk in a checkpoint.

docs/building-in-chunks/circuits.md
```diff
@@ -0,0 +1,178 @@
...
+## Estimated effort
+
+The design document for "building in chunks" estimates 5 weeks for chunking, without any circuit changes. We expect the following additional effort for the changes described here:
+
+- Updates to block header and new checkpoint header: 1-2 weeks
+- Changes to blob layout: 2-3 weeks
+- Updated circuit topology: 3-5 weeks
+- L1 changes: 1-2 weeks
+
+Total is 7-12 weeks, on top of the 5 weeks, for **12-17 weeks total**.
```
just-mitch, line 18: How parallelizable is this?

docs/building-in-chunks/circuits.md
```diff
@@ -0,0 +1,178 @@
+# Intro
+
+This is a design document that covers the changes needed in circuits, types, and potentially L1 contracts to support v3 of "building in chunks".
+
+The goal of this design is to promote _chunks_ into actual _blocks_, which get _checkpointed_ into L1 at regular intervals. This adds another level hierarchy in our block building: we now have multiple _chunks_, that are aggregated into _checkpoints_, which are aggregated into _epochs_. While we'd propose to refer to _chunks_ as _blocks_ directly, we'll keep using the term _chunk_ throughout this design document to avoid confusion with the current L2 blocks. But if we do adopt this design, an L2 _block_ will be what we list here as a _chunk_.
```
just-mitch, line 5: > This adds another level hierarchy in our block building

https://en.wikipedia.org/wiki/Fundamental_theorem_of_software_engineering

spalladino: @iAmMichaelConnor can confirm that this was exactly what I told him when discussing :-P

docs/building-in-chunks/circuits.md
```diff
@@ -0,0 +1,178 @@
...
+
+## Spec
+
+### Archive tree
+
+Given _chunks_ are the new _blocks_, the archive tree now contains _chunk header hashes_. This means that all changes to the state root happen in the _chunks_, and _checkpoints_ are just a way of grouping them together (just like epochs are).
+
+### Slots and timestamps
+
+We keep the current definition of L2 slot as the timestamp range reserved for a given proposer, which means that now a single L2 slot can have multiple chunks (blocks). All chunks within a given checkpoint will have the same timestamp.
```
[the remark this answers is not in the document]
spalladino: I'm not sure. I remember some potential issues in the old PoW Ethereum that were caused by miners having some discretion for changing the timestamp at will within a given range. If we enforce all chunks to have the same timestamp, we then know exactly the timestamp for each L2 slot, and there's no wiggle room.

docs/building-in-chunks/circuits.md
```diff
@@ -0,0 +1,178 @@
...
+  l1ToL2MessageTree: AppendOnlyTreeSnapshot
+  noteHashTree: AppendOnlyTreeSnapshot,
+  nullifierTree: AppendOnlyTreeSnapshot,
+  publicDataTree: AppendOnlyTreeSnapshot,
+}
+```
+
+#### Committing to `TxEffects`
+
+However, we still need a commitment to the contents of a chunk. Each `TxEffect` has several fields that are entered into the world state tree (nullifiers, note hashes, etc), so these are committed to as part of the chunk header state reference. But the `TxEffect` contains fields that are not committed into world state: logs and l2-to-l1 messages. So we still need to commit to them as part of the block hash. To do this, we can leverage the blob Poseidon sponge, which hashes all tx effects across a block, and include the start and end state of the sponge in the chunk header. This guarantees that the chunkhash is a commitment to all effects in that chunk.
```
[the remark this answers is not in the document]
spalladino: Yes, my mind was thinking in blocks and then trying to rename them to chunks as I was writing. I clearly failed.

docs/building-in-chunks/circuits.md
```diff
@@ -0,0 +1,178 @@
...
+
+```ts
+/** ChainId for the L2 block. */
+chainId: Fr;
+/** Version for the L2 block. */
+version: Fr;
+/** Slot number of the L2 block */
+slotNumber: Fr;
+/** Timestamp of the L2 block. */
+timestamp: UInt64;
```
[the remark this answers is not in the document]
spalladino: Let's keep the discussion on the comment above

PR #74, opened 2025-07-15
dd: Batch propose for building in chunks

spalladino:


docs/building-in-chunks/batch-propose.md
```diff
@@ -0,0 +1,73 @@
...
+
+If we land at lower TPS or less bytes per tx, this can lead to worse UX overall, since a proposer would be incentivized to only post a single block within their checkpoint, effectively increasing block times to 72s. If we assume 300 bytes per tx, which is the current lower bound for a transfer with out-of-band messaging (which should be the preference for app developers who want to reduce tx costs) at 6 TPS, a single blob takes the full 72s to be filled.
+
+On the other hand, if we consider 2400 bytes per tx effect, which is the size of a transfer with fpc support with all notes posted on-chain, then each blob can carry about 50 txs only. Assuming 10 TPS, this means we fill 2 blobs per L1 slot, so checkpoints cannot be longer than 3 L1 slots (36s).
```
joeandrews, line 21: Why is this? We have up to 9 blobs per block currently and this should increase by end of year to 12 - 18 when peerDas goes in.

docs/building-in-chunks/batch-propose.md
```diff
@@ -0,0 +1,73 @@
...
+This requires no circuit changes, and requires changes in the node related to building in chunks.
+
+## Numbers
+
+### Checkpoint and slot durations
+
+Since we are not making changes to circuits, each block still takes up at least one blob. If we target 6 blobs per tx, which may reduce our chances at L1 inclusion, we can squeeze no more than 6 slots per checkpoint. Assuming 1200 bytes per tx effect, each 128kb blob can carry about 100 txs. Assuming 10 TPS, we require 12kb per second for DA, which is 1 blob every 10.6 seconds, roughly 1 blob per L1 slot. This means the duration of a checkpoint cannot be more than 6 L1 slots (72s), and the duration of each L2 slot should be equivalent to one L1 slot (12s).
+
+Note that the above assumes 10TPS. If we go below this, then each block would not fully utilize its blob, leading to wasted blobspace and increased DA cost per tx. We should expect proposers to **not** produce a block for a slot if they have not filled the block, and instead accumulate txs into the following block in the slot within their checkpoint. This leads to less DA cost, which is borne by the proposer, but worse UX, since users now have to wait longer for L2 inclusion.
```
joeandrews, line 17: Can we not incentivise production via the block reward?

docs/building-in-chunks/batch-propose.md
joeandrews, line 21: Why is this so high? The private refund flow is not needed here. This should be max 3 notes, and one is transient I beleive.

docs/building-in-chunks/batch-propose.md
```diff
@@ -0,0 +1,73 @@
...
+
+If we land at lower TPS or less bytes per tx, this can lead to worse UX overall, since a proposer would be incentivized to only post a single block within their checkpoint, effectively increasing block times to 72s. If we assume 300 bytes per tx, which is the current lower bound for a transfer with out-of-band messaging (which should be the preference for app developers who want to reduce tx costs) at 6 TPS, a single blob takes the full 72s to be filled.
```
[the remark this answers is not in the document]
joeandrews: I also want to understand the costs per block more in the table as these seem very high.

docs/building-in-chunks/batch-propose.md
```diff
@@ -0,0 +1,73 @@
...
+
+| Action                         | Per block | Per batch | Gas cost |
+| ------------------------------ | --------- | --------- | -------- |
+| Base tx cost                   |           | x         | 21k      |
+| Post blob inputs               | x         |           | 20k      |
+| Setup epoch                    |           | x         | 3k       |
+| Update L1 gas fee oracle       |           | x         | 8k       |
+| Validate blob commitments      | x         |           | 8k       |
+| Compute base fee               |           | x         | 22k      |
+| Validate block header          | x         |           | 20k      |
```
joeandrews, line 41: Is this not also warm?

docs/building-in-chunks/batch-propose.md
```diff
@@ -0,0 +1,73 @@
...
+| Consume inbox                  | x         |           | 10k      |
+| Push messages to outbox        | x         |           | 9k       |
+| Emit log                       | x         |           | 3k       |
+| Unaccounted for                |           | x         | 20k      |
+
+The total gas cost from `propose` today is then 400k gas every 36s, or 11k gas per second. Under this new model, the cost per block is `20 + 8 + 20 + 22 + 30 + 10 + 9 + 3` or 122k gas, and the cost per batch is `21 + 3 + 8 + 22 + 200 + 3 + 20` or 277k gas. Assuming 72s checkpoints:
+
+- With 72s checkpoints and 12s blocks (best for UX), total gas cost per checkpoint is 1M, or 14k gas per second (worse than status quo)
+- Decreasing block frequency to once per 36s (same UX as today), total gas cost per checkpoint is 521k gas, or 7k gas per second (33% better than status quo)
+- Halving checkpoints to 36s to support 2400-byte txs, with one block every 6s (best for UX), total gas cost is 1M, or 28k gas per second (much worse than status quo)
```
joeandrews, line 55: I don't understand this. Why would checkpoints be every 36s here, and why is the data so high, what needs 4 notes and nullifiers with data on chain? At most a FPC for a private transaction needs 3 notes, and one is transient.

docs/building-in-chunks/batch-propose.md
```diff
@@ -0,0 +1,73 @@
...
+| Validate blob commitments      | x         |           | 8k       |
+| Compute base fee               |           | x         | 22k      |
+| Validate block header          | x         |           | 20k      |
+| Post and validate attestations |           | x         | 200k     |
+| Update pending chain tip       |           | x         | 3k       |
+| Store archive root             | x         |           | 22k      |
+| Store block log                | x         |           | 30k      |
+| Consume inbox                  | x         |           | 10k      |
+| Push messages to outbox        | x         |           | 9k       |
+| Emit log                       | x         |           | 3k       |
```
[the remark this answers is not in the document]
spalladino: We use logs for discovering new blocks being posted

docs/building-in-chunks/batch-propose.md
```diff
@@ -0,0 +1,73 @@
...
+| Base tx cost                   |           | x         | 21k      |
+| Post blob inputs               | x         |           | 20k      |
+| Setup epoch                    |           | x         | 3k       |
+| Update L1 gas fee oracle       |           | x         | 8k       |
+| Validate blob commitments      | x         |           | 8k       |
+| Compute base fee               |           | x         | 22k      |
+| Validate block header          | x         |           | 20k      |
+| Post and validate attestations |           | x         | 200k     |
+| Update pending chain tip       |           | x         | 3k       |
+| Store archive root             | x         |           | 22k      |
```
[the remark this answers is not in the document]
spalladino: Not the archive root. We keep the entire history of archive roots in storage (as opposed to other fields, which are kept in circular storage, that's why they are cheaper). This allows any client to easily detect if their current state is valid or not with a single CALL. We could probably replace this with circular storage at the expense of more complexity in the archiver.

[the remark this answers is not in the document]
spalladino: Also, a nit: cold/warm refers to whether the storage slot has been accessed at some point throughout the tx. I understand that here you meant whether the slot was zero initially or not.

docs/building-in-chunks/batch-propose.md
spalladino: > Why would checkpoints be every 36s here

This is assuming we consume 24kb/s in blob data and there are only 6 usable blobs per L1 blocks.

> why is the data so high,

Can I defer to you @iAmMichaelConnor on this?

LHerskind (commented):
Thanks for writing it here 👍 Not sure that I think it would take 8 weeks, but the benefit is too small anyway it seems from targets.

PR #83, opened 2025-09-26
Updated key store specification with BLS keys

PhilWindle:
This PR adds an optional specification for the inclusion of BLS keys.

### DIFF
--- a/docs/keys-config/prd.md
+++ b/docs/keys-config/prd.md
@@ -112,10 +112,7 @@ type BLSAccount =
   | JsonKeyFileV3Config;
 
 /** An AttesterAccount is a combined EthAccount and optional BLSAccount */
-type AttesterAccount = {
-  ethAccount: EthAccount;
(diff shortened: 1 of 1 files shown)


docs/keys-config/prd.md
```diff
@@ -52,6 +52,12 @@ type ValidatorKeyStore = {
 	 * An attester address may only appear once across all configuration blocks across all keystore files.
 	 */
 	attester: EthAccounts;
+	/** 
+	 * One or more BLS attester keys to handle in this configuration block.
+	 * These keys map 1 to 1 with the attester accounts above.
+	 * So for example, if a mnemonic is used here it should specify the same number of keys as in 'attester'.
+	 */
+	blsAttester?: BLSAccounts;
```
spalladino, line 60: Should we add a section on what happens if this is not set? Is the eth private key used to generate the bls one? Do we throw if the eth config uses remote signer and there is no corresponding bls entry?

PR #99, opened 2026-02-21
dd: proposer pipelining "build-ahead"

Maddiaa0:


iAmMichaelConnor (commented):
Great doc!

docs/build-ahead/dd.md
```diff
@@ -0,0 +1,554 @@
...
+| Dead zone | 24s (33%) | 12s (17%) |
+| Blocks per slot | 8 | 10 |
+| Latency improvement | — | ~12s |
+| Throughput improvement | — | +25% |
+
+See [ADR-001](#adr-001-conservative-trigger-attestation-gated).
+
+### Option 2: Speculative — checkpoint_proposal-Gated
+
+The next proposer starts building (locally, **not broadcasting**) as soon as the `checkpoint_proposal` is received. Blocks are only **broadcast after attestations arrive**.
```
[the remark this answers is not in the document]
Maddiaa0: ill combine the two sections together, theyre not different enough to warrant seperate sections

docs/build-ahead/dd.md
```diff
@@ -0,0 +1,446 @@
...
+
+With 6s blocks, the dead zone is at least **24 seconds out of every 72** — that's 33% of the slot with no block production. 8 blocks are produced per slot despite 72s of wall-clock time. The following analysis is based on 6s block durations.
+
+KPIs: we are trying to reduce **user-perceived latency** (time from TX submission to "proposed chain" visibility) by up to 12s and increase **effective chain throughput** from 8 blocks per slot to 10 blocks per slot (+25% improvement).
+
+The solution: allow the next slot's proposer to begin building blocks as soon as the predecessor's checkpoint data appears on P2P — before it lands on L1. The next proposer builds and broadcasts blocks during what is currently dead time, and anyone can submit the predecessor's checkpoint to L1 (the next proposer is incentivized to do so because their own blocks depend on it). All blocks B builds during the overlap go into B's checkpoint for B's slot.
```
PhilWindle: Is it not possible to stagger the slots entirely?

So, proposer A builds checkpoint 10 in slot 9 and publishes to L1 in slot 10. Meanwhile, proposer B builds checkpoint 11 in slot 10 (having seen checkpoint 10 on p2p towards the end of slot 9) and publishes in slot 11.

I guess everyone needs to agree to only attest to a checkpoint for slot N broadcasted in slot N - 1. Otherwise the proposer for slot N could just keep building all the way through slot N. But that should be ok.

docs/build-ahead/dd.md
```diff
@@ -0,0 +1,446 @@
...
+
+# Potential Improvement: L1 contract changes
+
+If proposer A misses their checkpoint submission slot (on purpose / being out priced), proposer B will have spent some time building on a chain that will be pruned, even though the attestation's gathered have given them a high confidence in their validity.
+
+If a proposer A misses their slot, I propose allowing it's submission for longer than the current deadline. Giving up to two l1 submission slots for each proposer. In this proposal, the proposerA must face a rather extreme penalty for missing their slot, either by a slashing, loosing a % of their rewards to the late proposal submitter - or both.
+
+## L1 Submission Handoff
+
+The predecessor attempts L1 submission during their slot as normal. At the **slot boundary**, the predecessor stops trying. After the slot boundary, anyone can submit the predecessor's checkpoint to L1 — the current `ProposeLib.sol` already allows any address to submit. The incentive for the next proposer to submit is indirect: B needs A's checkpoint on L1 to make B's own blocks valid, but also should yield some of proposer A's block rewards.
```
aminsammara: > but also should yield some of proposer A's block rewards.

This is on the roadmap. Data shows undeniably certain publishers are much better at landing checkpoints on l1 faster and cheaper. More longer term, I don't think proposers should submit anything to L1 and that should be handled by specialized publishers. 

Without this don't assume anyone is ever gonna submit someone else's block to L1. The current avg cost per block is 77% of the reward (assuming the increase in block rewards).

docs/build-ahead/dd.md
```diff
@@ -0,0 +1,446 @@
...
+
+When the overlap spans an epoch boundary (last slot of epoch N → first slot of epoch N+1), the committee rotates. The old committee's attestations **remain valid** — the L1 contract can verify signatures from the old committee regardless of rotation. No re-validation by the new committee is required.
+
+This means the overlap across epoch boundaries works identically to mid-epoch overlaps. The next proposer builds on the predecessor's attested checkpoint exactly as they would in any other slot transition.
+
+## Mid-Slot Liveness Timeout
+
+If the current proposer is unresponsive — no block proposals or checkpoint data appear on P2P — the network should not wait for the full slot to elapse. A **globally agreed, configurable timeout** fires mid-slot: if no proposals from the current proposer are observed by this deadline, honest nodes start accepting proposals from the next proposer.
+
+The next proposer builds from the **last confirmed state** (the most recent L1-confirmed tip). This is equivalent to the current system's behavior when a proposer is offline, but triggered earlier within the slot rather than waiting for the slot boundary.
```
aminsammara: > **last confirmed state** (the most recent L1-confirmed tip).

I don't think this is worth it. Any `m` less than the slotDuration leaves the proposer N+1 who builds early after the timeout trigger at risk of wasting work. 

Image

aminsammara: sorry for the confusing N+3 stuff but my point is as long as the L1 contract accepts proposals for the duration of the slot, having the next proposer build early could be a grieving vector. It gives us nothing in the happy pay (i.e. next proposer sees block proposals before `m`), in the unhappy path it still has shortcomings (what i describe above) and is probably a lot of work to implement.

docs/build-ahead/dd.md
```diff
@@ -0,0 +1,446 @@
...
+- **Depth limit**: Three consecutive proposers where the first two checkpoints don't land on L1 — third proposer should wait.
+- **Epoch boundary**: Overlap spanning committee rotation — old committee attestations accepted without re-validation.
+- **Same proposer**: Consecutive slots with the same proposer — overlap is skipped.
+- **Mid-slot liveness timeout**: Predecessor produces no proposals — liveness timeout fires, next proposer builds from confirmed state.
+- **Empty slot**: Predecessor offline — next proposer falls back to current behavior.
+- **Throughput**: Measure actual block count per slot with Build Ahead enabled vs disabled.
+
+# Rollout Plan
+
+This **requires** a coordinated rollout - creating a second proposer view will require all nodes to run compatible versions such that they are building along the correct tip.
```
aminsammara, line 288: It takes time for nodes to update. A significant % will not update for days and this will split the network's view of what the correct tip is.

docs/build-ahead/dd.md
Maddiaa0: Yes it is, the implementation plan will likely go this route. 

As long as the majority of the node software agrees to NOT attest to proposals that appear outside of the pipeline, then it should be safe. 

An area of risk is when we pass the agreed "timeout". If a proposer publishes their proposal late, then we need the majority of attestors to ignore it, otherwise they could potentially publish a proposal within the next pipelined slot, causing the new proposer to have to scramble. This case will need to be recoverable within the implementation as there may be an attestation "witholding" attack that could crash the node software if this was no covered.

docs/build-ahead/dd.md
Maddiaa0: yeah im a believer in decoupling these actions somewhat.

docs/build-ahead/dd.md
```diff
@@ -0,0 +1,554 @@
...
+- **Speculative block discards**: how often does the retry-then-discard path fire? This directly impacts user experience (soft reorgs).
+- **Overlap utilization**: how much of the dead zone is actually recovered? Measures the real-world latency gain.
+- **Build Ahead trigger**: how often does B trigger on checkpoint_proposal vs attestation vs fallback (early build)?
+- **Next-proposer L1 submissions**: how often does the next proposer land the predecessor's checkpoint vs the predecessor landing it themselves?
+- **Silent discard rate** (speculative mode only): how often does B build speculatively then discard because attestations didn't arrive?
+
+# Open Questions
+
+- **Reward economics:** Rewards are distributed at **proof arrival time**, not at checkpoint submission time. This means the submitter of the L1 checkpoint doesn't automatically get the reward — the proof associates the reward with the original proposer. The incentive for B to submit A's checkpoint is indirect (B needs A's checkpoint on L1 to make B's own blocks valid). The exact reward mechanism and incentive alignment needs formal analysis but is deferred from this design.
```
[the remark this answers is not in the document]
Maddiaa0: For implementation I have opted to run this without L1 changes. These can be part of a sep proposal to improve submission across the board. Nothing should be specific to this implementation

spalladino (approved):

